PR #5, opened 2023-03-24
adding mac os x binaries to virtualenv setup

rbanick:


clemp (approved):
Excellent. This is what I had to do on my MacOS and then I was able to run the code.

PR #8, opened 2023-05-01
add host parameter to dataset_manager to make the ipfs host configurable

eschechter:


rbanick (approved):
Would be nice to add type hinting directly to the args in `init` while we’re at it, but not essential

PR #11, opened 2023-05-03
fix: broken links

Faolain:


clemp (commented):
So actually this link points to a non-existing branch on this repo. The correct link should be `pip install git+https://github.com/Arbol-Project/zarr-climate-etl-ipfs`

rbanick (approved):

Issue #20, opened 2023-05-15
Subprocess `Popen` sometimes blocking resources and crashing ETLs


rbanick:
After instituting the new `Popen` loop for transforming files I observed the following error during the transform step of a large ETL (15K files).

```
Encountered exception during execution:
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/prefect/engine.py", line 1581, in orchestrate_task_run
    result = await call.aresult()
  File "/usr/local/lib/python3.10/dist-packages/prefect/_internal/concurrency/calls.py", line 181, in aresult
    return await asyncio.wrap_future(self.future)
  File "/usr/local/lib/python3.10/dist-packages/prefect/_internal/concurrency/calls.py", line 194, in _run_sync
    result = self.fn(*self.args, **self.kwargs)
  File "/etls/prefect/etl_methods.py", line 26, in transform
    raise e
  File "/etls/prefect/etl_methods.py", line 23, in transform
    manager.transform()
  File "/usr/local/lib/python3.10/dist-packages/zarr_etl_tools/dataset_manager.py", line 197, in transform
    self.prepare_input_files()
  File "/usr/local/lib/python3.10/dist-packages/zarr_etl_managers/ETL.py", line 323, in prepare_input_files
    self.fils_to_nc4s(input_dir, keep_originals)
  File "/usr/local/lib/python3.10/dist-packages/zarr_etl_managers/prism.py", line 342, in fils_to_nc4s
    self.parallel_subprocess_files(fils, command_text, '.nc', keep_originals)
  File "/usr/local/lib/python3.10/dist-packages/zarr_etl_tools/utils/zarr_methods.py", line 206, in parallel_subprocess_files
    processes = [ Popen(command) for command in commands ]
  File "/usr/local/lib/python3.10/dist-packages/zarr_etl_tools/utils/", line 206, in <listcomp>
    processes = [ Popen(command) for command in commands ]
  File "/usr/lib/python3.10/subprocess.py", line 969, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.10/subprocess.py", line 1778, in _execute_child
    self.pid = _posixsubprocess.fork_exec(
BlockingIOError: [Errno 11] Resource temporarily unavailable
```

This crashes the ETL in question. 

To investigate
* What causes the issue?
* Can it be resolved just by waiting, and if so, how to wait?
[...]

PR #21, opened 2023-05-15
running 100 Popen processes at a time instead of all simultaneously

rbanick:
Addressing #20 by only passing 100 subprocess commands at a time.

This passes tests locally


eschechter (commented):
I don't think this is correct. They're still all going to run at once with this - Popen is what triggers the subprocess. `.wait()` just says to not continue until the subprocess completes.

eschechter (approved):

PR #30, opened 2023-05-22
Select dataset by name

rbanick:


rbanick:
Tested in Prefect prod and it works

gridded_etl_tools/dataset_manager.py
```diff
@@ -78,6 +79,9 @@ def __init__(
         global_log_level : str, optional
             The root logger `logger.getLogger()` will be set to this level. Recommended to be `logging.DEBUG`, so all logging
             statements will be generated and then logging handlers can decide what to do with them.
+        dataset_name : str, optional
```
ohsqueezy, line 82: I think we're actually dropping this

gridded_etl_tools/dataset_manager.py
```diff
@@ -453,7 +455,7 @@ def run_etl(
 
         """
         # Find the dataset class (e.g. CHIRPSPrelim05) from its name string
-        dataset_class = self.get_dataset_manager_from_name(dataset_name)
+        dataset_class = self.get_subclass(dataset_name)
```
ohsqueezy, line 458: This can be for a later PR, but now that we're getting around to cleaning up this area of the class, I think if we even keep run_etl, we want it to either only run its own class or to become a classmethod like get_subclass

ohsqueezy (commented):
I think we want to drop `dataset_name` from the init function, but otherwise looks good

gridded_etl_tools/dataset_manager.py
rbanick: Definitely for a later PR, but I also agree.

I suspect we might end up ditching `run_etl` however.

gridded_etl_tools/dataset_manager.py
rbanick: will amend

ohsqueezy (approved):

PR #31, opened 2023-05-22
Default s3 authentication

ohsqueezy:
s3fs will check for environment variables automatically, so we can just initialize S3FileSystem using default initialization and let the library handle authentication. Clarified the docs and logs to indicate that authentication is not performed at initialization.


ohsqueezy:
Tested with gridded-etl-managers by checking out this branch and running the pytests. The version string is edited because installing the branch through pip isn't working without updating the version.

rbanick (approved):

PR #33, opened 2023-05-25
improving docstrings

rbanick:


gridded_etl_tools/utils/convenience.py
```diff
@@ -631,9 +632,9 @@ def check_if_new_data(self, existing_file_end_date: datetime.datetime) -> bool:
...
             return False
         self.info(f"newest file ends at {newest_file_end_date}")
-        if newest_file_end_date >= existing_file_end_date:
-            self.info(f"newest file has newer data than our end date {existing_file_end_date}, triggering parse")
+        if newest_file_end_date >= compare_date:
+            self.info(f"newest file has newer data than our end date {compare_date}, triggering parse")
             return True
         else:
-            self.info(f"newest file doesn't have data past our existing end date {existing_file_end_date}")
+            self.info(f"newest file doesn't have data past our existing end date {compare_date}. Exiting script.")
```
ohsqueezy, line 639: We don't want to log that it's exiting the script because callers are expected to be able use this function arbitrarily

ohsqueezy (changes requested):

ohsqueezy (approved):

PR #38, opened 2023-06-12
make thread count a factor of RAM size

ohsqueezy:
Non-urgent proposed improvement to the calculation of number of threads to use. Currently it is a factor of the number of CPUs, but this would make it a factor of RAM. Comes with some unit tests which test the following ratios.

```
INFO     <chirps_final_25> Using 24 threads on a 32-core system with 256.00GB RAM
INFO     <chirps_final_25> Using 12 threads on a 32-core system with 128.00GB RAM
INFO     <chirps_final_25> Using 16 threads on a 16-core system with 256.00GB RAM
INFO     <chirps_final_25> Using 1 threads on a 8-core system with 8.00GB RAM
INFO     <chirps_final_25> Using 1 threads on a 1-core system with 32.00GB RAM
```

### DIFF
--- a/gridded_etl_tools/dataset_manager.py
+++ b/gridded_etl_tools/dataset_manager.py
@@ -7,6 +7,7 @@ import multiprocessing
 import multiprocessing.pool
 import argparse
 import datetime
+import psutil
 
 from .utils.logging import Logging
 from .utils.zarr_methods import Publish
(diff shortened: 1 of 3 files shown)


rbanick (approved):
I like it. Curiosity question, why the ratio of 3/32? Is this recommended somewhere

ohsqueezy:
That was the ratio we used for ERA5 full history, 256RAM to 24 cores. Even though there were 32 cores, using 24 seemed to work better. I think we also ran an ETL with 12 cores on a 128GB/16 core system that worked well.

The ratio might make sense to increase as the RAM size gets smaller, like if we run ETLs on a 8GB RAM system, we might want it to be using more than one thread.

PR #43, opened 2023-07-20
adding ability to backdate ('pre-pend') to existing datasets without …

rbanick:
…incident


rbanick:
Closing PR b/c, while the code works, it causes a horribly inefficient rechunking operation. Easier just to re-parse.

PR #44, opened 2023-07-25
add optional file path to zarr_json_to_dataset

rbanick:
Adding an optional parameter specifying the Zarr JSON being opened. Useful for testing and developing.


ohsqueezy (approved):
Looks OK to merge as is but in general we should try to keep the docstrings updated when new parameters are added, especially in this case because it's an optional parameter with the same name as a member function so the docstring can make it clear when each gets used

PR #47, opened 2023-07-31
use typing-only import for dm in store module

ohsqueezy:
Since the store module is only using DM for typing checks, we can use the [typing-only annotations import](https://peps.python.org/pep-0563/) and remove the DM module import. This is passing tests in the managers lib as well.


ohsqueezy:
Working on getting the linter to accept this

ohsqueezy:
New version uses both the annotations dict and the TYPE_CHECKING indicator and passes all tests and the linter (my understanding was using both isn't necessary in Python 3.10 but this is the only way that is working). I also confirmed Evan's original bug and checked that this fixes it.

rbanick (approved):

PR #48, opened 2023-08-07
Date range error

rbanick:
Minor error fix for date_range check


gridded_etl_tools/dataset_manager.py
```diff
@@ -196,11 +196,13 @@ def static_metadata(self):
         ...
 
     @abstractmethod
-    def extract(self):
+    def extract(self, date_range: tuple[datetime.datetime, datetime.datetime] = None):
```
ohsqueezy, line 199: since the argument can be `None` the typing should use `typing.Optional`

ohsqueezy (approved):

PR #54, opened 2023-08-09
dataset specific properties

rbanick:


gridded_etl_tools/dataset_manager.py
```diff
@@ -56,7 +56,9 @@ def __init__(
         allow_overwrite=False,
         ipfs_host="http://127.0.0.1:5001",
         dask_dashboard_address: str = "127.0.0.1:8787",
-        forecast: bool = False,
+        write_local_zarr_jsons: bool = False,
```
ohsqueezy, line 59: Since we're implementing this just in GFS for now, maybe we want to drop it from the argument list

ohsqueezy (approved):

PR #57, opened 2023-08-14
rewiring zarr_json_path to export w/in datasets

rbanick:
Finding the merged MultiZarr for ETLs run on a remote server is tricky as it gets buried in the root directory, which can be a surprisingly ambiguous location on containerized systems. I’d prefer to consolidate its placement alongside downloaded files for ease of reference.


ohsqueezy (approved):

rbanick:
You’re right, but unfortunately it’s not so simple.

The problem I was not correctly surfacing is that `populate_metadata` is w/in Transform and Parses will fail without running that. So `skip_transform` as a standalone argument actually never works.

Therefore I think we need to institute a `use_existing_zarr_json` parameter that’s passed during orchestration. It will default to False so that for daily updates and other normal operating procedures the file is overwritten.

PR #67, opened 2023-08-31
feat: encryption

chrisrossi:
Add an easy to use encryption filter for reading/writing Zarrs. Encryption uses XChaCha20-Poly1305.


gridded_etl_tools/utils/metadata.py
```diff
@@ -683,6 +684,13 @@ def encode_vars(self, dataset: xr.Dataset) -> xr.Dataset:
                         "units": None,
                     }
                 )
+
+        # Encrypt variable data if requested
+        if self.encryption_key is not None:
+            dataset[self.data_var()].encoding["filters"] = [
+                EncryptionFilter(self.encryption_key)
+            ]
```
eschechter, line 692: Nitpicking a bit, but I think we should check for existing filters and append if there are any instead of always overwriting. Maybe not strictly necessary for code in current state, but helps future proof.

gridded_etl_tools/dataset_manager.py
```diff
@@ -9,6 +9,7 @@
 import datetime
 import psutil
 
+from .encryption import register_encryption_key
```
eschechter, line 12: `encryption.py` should move into the `utils` folder in keeping with how we handle other helper modules.

gridded_etl_tools/encryption.py
```diff
@@ -0,0 +1,138 @@
...
+    ----------
+
+    key_hash: str
+        The hex digest of an encryption key. A key may be generated using
+        :func:`generate_encryption_key`. The hex digest is obtained by registering the
+        key using :func:`register_encryption_key`.
+    """
+
+    codec_id = "xchacha20poly1305"
+    header = b"dClimate-Zarr"
```
eschechter, line 96: Should `header` be customizable? We want the package to not be tied to dClimate too closely.

gridded_etl_tools/dataset_manager.py
chrisrossi: 055d4ab

The reason I initially had it the level I did was it contains one function I consider part of the public API (`generate_encryption_key`) and I was trying to keep that closer to the surface. It wasn't a great solution, though. I think it might be helpful, in a future PR, to think about what the public API for the library is and re-export anything  we expect the user to interact with directly in the root of the package and focus our documentation there.

gridded_etl_tools/utils/metadata.py
chrisrossi: 0d581f0

True. Currently, we know that `remove_unwanted_fields` is going to initialize that for us, but best not to assume.

gridded_etl_tools/encryption.py
chrisrossi: This is not user facing and it is completely arbitrary. AFAICT, it's function is to help verify the decryption by being a known value that can be checked for, and could be literally anything. We could always change it something very generic looking, but I don't think there's any need to make it customizable.

eschechter (approved):

PR #68, opened 2023-09-07
feat: enable compression for all stores

chrisrossi:
Based on real world experimentation, I'm recommending we use compression for all data stores.

To get some real world numbers, I first wrote a [script to write 10 years worth of CHIRPS data to a Zarr](https://gist.github.com/chrisrossi/f44d8416a4ef802209df901ba575535d). I ran it once without compression and then ran it again with compression, so I'd have two datasets to compare. I then ran a [script to calculate space saving from IPLD deduplication](https://gist.github.com/chrisrossi/ce5eb9a3d65d01c406cd44286c0b74e5) on each dataset.

```
(zarr) chris@spirit:~/proj/arbol/gridded-etl-tools$ python analyze_duplicates.py `cat uncompressed_cid`
IPFS Deduplication has saved 847,384,732/5,300,734,082 bytes (15.99%)
(zarr) chris@spirit:~/proj/arbol/gridded-etl-tools$ python analyze_duplicates.py `cat compressed_cid`
IPFS Deduplication has saved 18,488,975/1,265,650,096 bytes (1.46%)
```

The results are, then: 

| Data | Size |
| ---------------- | --------------- |
| Without compression or deduplication | 5,300,734,082 (5.30G) |
| Without compression, with deduplication from IPLD |  4,453,349,350 (4.45G) |
| With compression (and a little deduplication from IPLD) | 1,247,161,121 (1.25G) |

In this sample, at least, compression saves a lot more disk space than IPLD deduplication, so there's no need to refrain from compressing in the hopes that IPLD deduplication will be more helpful. I have no reason to think the results will be vastly different with a different dataset, but we can certainly try some more if we want to gather more data.

### DIFF
--- a/gridded_etl_tools/dataset_manager.py
+++ b/gridded_etl_tools/dataset_manager.py
@@ -62,6 +62,7 @@ class DatasetManager(Logging, Publish, ABC, IPFS):
         use_local_zarr_jsons: bool = False,
         skip_prepare_input_files: bool = False,
         encryption_key: str = None,
+        use_compression: bool = True,
         *args,
         **kwargs,
     ):
@@ -104,6 +105,8 @@ class DatasetManager(Logging, Publish, ABC, IPFS):
             If provided, data will be encrypted using `encryption_key` with
             XChaCha20Poly1305. Use :func:`.encryption.generate_encryption_key` to
             generate a random encryption key to be passed in here.
+        use_compression: bool, optional
+            Data in this dataset will be compressed unless this is set to `False`.
         """
         # call IPFS init
         super().__init__(host=ipfs_host)
@@ -177,6 +180,8 @@ class DatasetManager(Logging, Publish, ABC, IPFS):
         self.info(f"Using {self.dask_num_threads} threads on a {multiprocessing.cpu_count()}-core system with {total_memory_gb:.2f}GB RAM")
 
         self.encryption_key = register_encryption_key(encryption_key) if encryption_key else None
+
+        self.use_compression = use_compression
 
     # SETUP
 
--- a/gridded_etl_tools/utils/metadata.py
+++ b/gridded_etl_tools/utils/metadata.py
@@ -801,11 +801,7 @@ class Metadata(Convenience, IPFS):
             The dataset being published, after metadata update
 
         """
-        # Leave compression off for IPLD
-        if isinstance(self.store, IPLD):
-            compressor = None
-        else:
-            compressor = numcodecs.Blosc()
+        compressor = numcodecs.Blosc()
 
         for coord in ["latitude", "longitude"]:
             dataset[coord].attrs.pop("chunks", None)
--- a/gridded_etl_tools/utils/metadata.py
+++ b/gridded_etl_tools/utils/metadata.py
@@ -801,7 +801,7 @@ class Metadata(Convenience, IPFS):
             The dataset being published, after metadata update
 
         """
-        compressor = numcodecs.Blosc()
+        compressor = numcodecs.Blosc() if self.use_compression else None
 
         for coord in ["latitude", "longitude"]:
             dataset[coord].attrs.pop("chunks", None)
--- a/tests/utils/test_metadata.py
+++ b/tests/utils/test_metadata.py
@@ -1,7 +1,10 @@
 from unittest import mock
 
+import numcodecs
+
 from gridded_etl_tools import dataset_manager
 from gridded_etl_tools.utils import encryption
+from gridded_etl_tools.utils import store
 
 
 def unimplemented(*args, **kwargs):
@@ -73,3 +76,11 @@ class TestMetadata:
         assert len(filters) == 2
         assert filters[0] == "SomeOtherFilter"
         assert isinstance(filters[1], encryption.EncryptionFilter)
+
+    def test_remove_unwanted_fields_w_ipld_store(self):
+        dataset = mock.MagicMock()
+        dataset["data"].encoding = {}
+        md = DummyManager()
+        md.store = store.IPLD(md)
+        dataset = md.remove_unwanted_fields(dataset)
+        assert isinstance(dataset["data"].encoding["compressor"], numcodecs.Blosc)
--- a/tests/utils/test_metadata.py
+++ b/tests/utils/test_metadata.py
@@ -84,3 +84,11 @@ class TestMetadata:
         md.store = store.IPLD(md)
         dataset = md.remove_unwanted_fields(dataset)
         assert isinstance(dataset["data"].encoding["compressor"], numcodecs.Blosc)
+
+    def test_remove_unwanted_fields_w_ipld_store_no_compression(self):
+        dataset = mock.MagicMock()
+        dataset["data"].encoding = {}
+        md = DummyManager(use_compression=False)
+        md.store = store.IPLD(md)
+        dataset = md.remove_unwanted_fields(dataset)
+        assert dataset["data"].encoding["compressor"] is None


ohsqueezy:
It's been a while since we last talked about compression on IPLD, but if I'm remembering correctly the issue with turning it on is that it would break updates to existing datasets because we don't have a solution for appending new compressed data to a compressed dataset.

Last we discussed early in the year, we were looking at two possible solutions for appending compressed data

* Recursive hashes at the end of the Zarr
* Variable chunk lengths feature of Zarr 3.0

chrisrossi:
> It's been a while since we last talked about compression on IPLD ...

I guess your concern is that on appends you may end up with more stale blocks than you might otherwise? I think if you just unpin outdated blocks and garbage collect them, you'll find it doesn't make much difference in the long term. But maybe I don't understand the whole problem. Is there a place in Zarr documentation or your code where I can see what's currently being done for appends to try and minimize the amount of garbage produced?

ohsqueezy:
I realized I should have mentioned that currently we don't unpin previous hashes because we keep a reference to each dataset's previous hash in the metadata as we update. I think the practical application is that if there is ever a dispute over what data a contract was signed on, the hash used in the contract will still return the data.

I think there are still ways forward with compression, though. These are two ideas we've discussed, although neither of them have gotten to the proof of concept stage.

- Currently, with Zarr v2, all chunks need to be the same size. With variable chunk lengths in Zarr v3, we wouldn't be locked into recomputing the hash of the final chunk of data. Theoretically, we could just append a new chunk every update. I'm not sure if we would then run into an issue of having too many chunks, though.
- If hashes at the end of the Zarr were recursive hashes, each one could could have a new hash appended to it, and then just the root of the recursive hash would change, but no data chunks would need to be recomputed. I guess one thing to test with this would be to see if using a recursive hash to refer to the data chunk is possible.

chrisrossi:
Hi, 

So I've done a little bit of experimentation just to see where things stand currently. For this experiment, I created a Chirps dataset with data from 2012-2022. I then added data through 8/2023. I did this both with and without compression, resulting in 4 trees that can be examined: a before and after without compression and a before and after with compression. I then wrote a [script to compare two trees](https://gist.github.com/chrisrossi/d103d5ac52c1c9bb49faeaf249109dad) and find the leaf node blocks that are in one tree and not the other. Using this to compare the before/after trees, we can figure out how much data is left in blocks that aren't used in the new tree after the append. 

For uncompressed data, this is the result:

~~~
(zarr) chris@spirit:~/proj/arbol/gridded-etl-tools$ python diff_trees.py `cat uncompressed_cid` `cat uncompressed_after`
0 blocks discarded (0 bytes)
42070 blocks added (269,704,938 bytes)
~~~

Basically, there are no blocks that are in the before tree that aren't also in the after tree, so there is no waste. That is impressive! Next, I suppose I'll be reading through the `ipldstore` code to see how that's done.

For compressed data, this is the result:

~~~
(zarr) chris@spirit:~/proj/arbol/gridded-etl-tools$ python diff_trees.py `cat compressed_cid` `cat compressed_after`
1307 blocks discarded (7,643,130 bytes)
12864 blocks added (81,832,962 bytes)
~~~

It looks like the compression did cause some blocks to become stale when appending. Look at the number of bytes we're, talking about though. In the uncompressed case, the append added ~270mb to the size of the dataset. In the compressed case, we got ~82mb of new data plus about 7.6mb of stale data. Combining those numbers, though, we get ~89mb which is around a third of 270mb, so even factoring in the inefficiency of creating some stale blocks with compression, compression still "wins" at keeping the total dataset size down.

This experiment can be done with no knowledge of how appends are handled by `ipldstore` and just looks at the the size of the dataset before and after an append. It may be that after becoming more familiar with `ipldstore` and understanding how stale blocks are avoided in the uncompressed case, some path will become apparent to make appends more efficient with compression. Even without that, though, there still seems to be a net near term gain realized by simply turning compression on now.

chrisrossi:
So, I've taken a look at a larger dataset with more frequent updates. Specifically, I've loaded a year's worth of data for ERA5 precipitation and then appended 32 more days worth of data, 24 hours at a time, for a total of 33 datasets. I've done this once with compression and once without compression, to run an analysis of how the dataset evolves with each append in both scenarios. It turns out the salient points can be seen just by looking at the first append to the year-long dataset.

```
$ ipfs dag stat --progress=false `cat era5-precip-2021-12-31`  # commas added for clarity

CID                                                             Blocks          Size
bafyreiaqlnneyknmbeqbl2qh2umynvtfrktiyseytcdocsonexupo2o7fi     1453837         35,257,870,108

Summary
Total Size: 35,257,870,108
Unique Blocks: 1453837
Shared Size: 0
Ratio: 1.000000

$ ipfs dag stat --progress=false `cat era5-precip-compressed-2021-12-31`

CID                                                             Blocks          Size
bafyreiaa7ahrpyxfvdnxp6tbto656t4z662yie6fikwddrp2bfuxa2nnnq     296431          7,088,311,219

Summary
Total Size: 7,088,311,219
Unique Blocks: 296431
Shared Size: 0
Ratio: 1.000000

$ python diff_trees.py `cat era5-precip-compressed-2021-12-31` `cat era5-precip-compressed-2022-01-01`
8090 blocks discarded (175,821,369 bytes)
10485 blocks added (218,177,175 bytes)
$ python diff_trees.py `cat era5-precip-2021-12-31` `cat era5-precip-2022-01-01`
3522 blocks discarded (86,605,980 bytes)
7219 blocks added (177,490,714 bytes)
```

We can see here that even in the uncompressed dataset there is some waste generated by appending data. (Maybe 24 hours isn't the update interval?) We do also see that the compressed dataset produces more waste than the uncompressed dataset. While the compressed dataset is starting at at about 20% the size of the uncompressed dataset, each append produces around twice the waste. How many appends it would take to swamp out the effects of the compression depends on the initial size of the dataset. For one year's worth of data, the uncompressed data is 35gb while the compressed data is 7gb, for a savings of 28gb. If we start the dataset off with, say, 40 years worth of data (dClimate's data starts Jan 1, 1979, so we're currently at 44 years and counting), than compression saves us 40 * 28 = 1120gb. For each append, the uncompressed data produces 86mb of waste, while the compressed dataset produces about 176mb, for a difference of about 90mb. 1120gb  / 90 kb = 12444 days or about 34 years before the space savings of compression is negated by the added waste of appends with compression.

Of course, this assumes that we retain all the history of every append over 34 years. With careful pruning of the tree (perhaps we only retain datasets with contracts) 34 years can become never. Also, assuming this technology will still be in use for this purpose 34 years from now is optimistic, to say the least. (Or pessimistic, from the point of view of nothing better comes along in that time to supersede it.) 

All that said, I still think compression at the Zarr level is the way to go, and presents an exceedingly easy path forward to saving considerable amounts of money on storage.

There will, of course, be a performance penalty, as yet unquantified, due to the extra step of decompressing data when querying the dataset. With some proper benchmarking we could put some numbers to that, if that was seen as worthwhile. 

It is also possible, but harder since it hasn't already been implemented for us, to add compression at the ipldstore layer. This gets around the issue of wasteful appends by making the block boundaries identical to the uncompressed dataset, just with smaller blocks in IPFS. The only real downside to this approach, aside from the effort to implement, which probably isn't too bad, honestly, is if you're compressing at the Zarr layer you can keep your IPFS blocks at about the same size, and just store fewer blocks in IPFS, which cuts down on the number of DHT lookups you do in IPFS to retreive the whole dataset. Just compressing the blocks at the IPFS layer is a little less efficient as you're not utilizing all the space available in individual blocks in IPFS. Again, some benchmarking could put numbers to that. I'm happy to implement this if you all would like.

My recommendation, though, remains to just use the already implemented compression at the Zarr layer.

ohsqueezy:
Awesome, I actually wanted to run exactly this scenario using your script. I think this is more than enough to show compression is usable as is for all of the use cases I've seen at least.

Maybe this is too cautious or impractical, but I would still consider having it optional. I think a reasonable use case for the library is with datasets that have little or no history, and in that case wouldn't the waste immediately be larger with compression? And I'm not super familiar with the range of types of data stored on Zarr, but there could also be a case where the data doesn't compress much.

These are also solved by not keeping the previous hash, so I also think making storing the previous hash optional would be a good addition to go along with this.

chrisrossi:
I can definitely make the compression thing configurable. I'll look into how easy/hard it is to make history retention configurable. That might be a separate PR.

rbanick (approved):

PR #71, opened 2023-09-25
Bumping versions of Zarr, Xarray, and related libraries. Dask stays at 2023.3.0

rbanick:


setup.cfg
```diff
@@ -15,9 +15,9 @@ python_requires = >=3.10
 install_requires =
     cfgrib
     dag-cbor==0.2.2
-    dask[array,diagnostics,distributed]==2023.1.1
+    dask[array,diagnostics,distributed]==2023.3.0
```
eschechter, line 18: make it `< 2023.4.0` to give pip some leeway

gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -412,45 +413,44 @@ def parse(self, *args, **kwargs) -> bool:
...
-                    self.update_zarr()
-                elif not self.store.has_existing or (
-                    self.rebuild_requested and self.overwrite_allowed
-                ):
-                    if not self.store.has_existing:
-                        self.info(
-                            f"No existing data found. Creating new Zarr at {self.store}."
-                        )
+        ) as cluster:
+            with Client(cluster) if not isinstance(self.store, IPLD) else nullcontext() as client:
```
eschechter, line 424: Do we want to make the nullcontext change? Or does it not matter because we're sticking with older dask

gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -412,45 +413,44 @@ def parse(self, *args, **kwargs) -> bool:
...
+                except KeyboardInterrupt:
+                    self.info(
+                        "CTRL-C Keyboard Interrupt detected, exiting Dask client before script terminates"
                     )
-            except KeyboardInterrupt:
-                self.info(
-                    "CTRL-C Keyboard Interrupt detected, exiting Dask client before script terminates"
-                )
-                client.close()
+                    client.close()
```
eschechter, line 453: If we stick with nullcontext, we need to remove this line, since nullcontext has no close method. In general though, I don't think this line does anything, since exiting the Client context forces it to close anyways

gridded_etl_tools/utils/zarr_methods.py
rbanick: Let’s keep it in there since it took us a while to figure out?

rbanick: Also since we’re pretty sure IPLD objects probably shouldn’t have pickled successfully in prior versions of Dask

eschechter (approved):

PR #76, opened 2023-10-04
Export Kerchunked Zarr JSONs

rbanick:
- implementing method to export Zarr JSONs within Kerchunkify
- adding forecast and ensemble hour default properties


gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -78,7 +79,7 @@ def create_zarr_json(self, force_overwrite: bool = True):
         else:
             self.info("Existing Zarr found, using that")
 
-    def kerchunkify(self, file_path: str, scan_indices: int = 0):
+    def kerchunkify(self, file_path: str, scan_indices: int = 0, local_file_path: str = None):
```
ohsqueezy, line 82: type hint should be `Optional[str]` since None is allowed

gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -131,9 +134,28 @@ def kerchunkify(self, file_path: str, scan_indices: int = 0):
...
+                self.zarr_json_in_memory_to_file(scanned_zarr_json, local_file_path)
 
         return scanned_zarr_json
 
+    def zarr_json_in_memory_to_file(self, scanned_zarr_json: str, local_file_path: str):
+        """
+        Export a Kerchunked Zarr JSON to file.
+        Implementations of this method rely on 
+        """
+        if hasattr(self, "data_driven_zarr_json_name"):
```
ohsqueezy, line 152: My feeling is we want to avoid hasattr as much as possible. In this case, I think having a base class `DatasetManager.data_driven_zarr_json_name` function that just returns its input and then overriding it in CFS would be a way to do the same thing without needing hasattr.

ohsqueezy:
Regarding `DatasetManager.write_local_zarr_jsons`, I think there are a few things we could do

- It seems like right now the local_file_path is always being passed to kerchunkify even though it's optional. Could we simply always write local zarr JSONs, or would that take up too much space?

If we want to keep the option

- Could we have a @property `DatasetManager.local_zarr_json_path` that is `None` by default? Then it becomes both the option and path itself.
- If that doesn't make sense either, I think a check in kerchunkify that asserts that local_file_path is not None when self.write_local_zarr_json is True would be necessary to add

ohsqueezy (changes requested):
Left a couple of comments and some thoughts about write_local_zarr_json above

rbanick:
> Regarding `DatasetManager.write_local_zarr_jsons`, I think there are a few things we could do
> 
> * It seems like right now the local_file_path is always being passed to kerchunkify even though it's optional. Could we simply always write local zarr JSONs, or would that take up too much space?

We don’t pass `local_file_path` when using Kerchunk w/in the Transform step on non-S3 based datasets that we download and process locally.

There’s no strong reason not to always write local Zarr JSONs, but neither do I think it’s optimal. It adds unnecessary overhead.

> 
> If we want to keep the option
> 
> * Could we have a @Property `DatasetManager.local_zarr_json_path` that is `None` by default? Then it becomes both the option and path itself.
> * If that doesn't make sense either, I think a check in kerchunkify that asserts that local_file_path is not None when self.write_local_zarr_json is True would be necessary to add

The Zarr JSON path will vary per day/forecast/ensemble so I don’t think a property makes sense.
I’m OK with a check.

gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -146,15 +149,36 @@ def kerchunkify(self, file_path: str, scan_indices: int = 0, local_file_path: st
...
+        If necessary, create a file name for that JSON in situ based on its attributes.
         """
-        if hasattr(self, "data_driven_zarr_json_name"):
-            local_file_path = self.data_driven_zarr_json_name(scanned_zarr_json)
+        local_file_path = self.file_path_from_zarr_json_attrs(scanned_zarr_json=scanned_zarr_json, local_file_path=local_file_path)
         with open(local_file_path, "w") as file:
             json.dump(scanned_zarr_json, file, sort_keys=False, indent=4)
             self.info(f"Wrote local JSON to {local_file_path}")
 
+    def file_path_from_zarr_json_attrs(scanned_zarr_json: dict, local_file_path: str) -> str:
```
ohsqueezy, line 160: We probably want to use pathlib.Path instead of str

ohsqueezy (approved):
The docstring is looking really solid. I think this can be merged although I would consider using pathlib instead of str.

PR #81, opened 2023-10-18
Tests for (some) Zarr methods

rbanick:
Adding unit tests for some critical Zarr methods.

This work was interrupted by urgent requests elsewhere so it only partially addresses our testing coverage gap. Still, it’s a good start and provides a foundation and workflow for additional work.


chrisrossi (approved):
LGTM. As long as tests pass (and they do), I'm happy, as I'll be in there mucking about anyway.

PR #85, opened 2023-10-30
Ensemble Forecasts and Hindcasts

rbanick:
* Introduce structure for hindcasts
* Changes needed to flexibly use `create_zarr_json` outside of normal transformation workflow
* Allow for alternative methods of preparing the final, transformed dataset than `zarr_json_to_dataset`
* Modularization of key transformation methods
* Additional toggles and triggers
* Refactoring code.


gridded_etl_tools/utils/attributes.py
```diff
@@ -181,6 +187,19 @@ def ensemble_numbers(self) -> list[int]:
         """To be overwritten by actual ensemble datasets"""
         return list(None)
 
+    @property
+    def hindcast_steps(self) -> list[int]:
+        """To be overwritten by actual hindcast datasets"""
+        return list(None)
+
+    @classmethod
+    def irregular_update_cadence(self) -> bool | tuple[np.timedelta64, np.timedelta64]:
```
ohsqueezy, line 196: Is it necessary to return False or could this return None instead? It's more common to use Optional[..] than to either return a bool or a value.

gridded_etl_tools/dataset_manager.py
```diff
@@ -58,6 +58,8 @@ def __init__(
         allow_overwrite=False,
         ipfs_host="http://127.0.0.1:5001",
         dask_dashboard_address: str = "127.0.0.1:8787",
+        dask_worker_memory_target: float = 0.65,
```
ohsqueezy, line 61: We should really start documenting these params in the docstring but understand if you want to leave it for a future PR

gridded_etl_tools/utils/convenience.py
```diff
@@ -664,4 +661,25 @@ def standardize_longitudes(cls, dataset: xr.Dataset) -> xr.Dataset:
...
+    @classmethod
+    def span_to_timedelta(cls) -> np.timedelta64:
+        """
+        Map a dataset's string time span to a corresponding numpy timedelta64 object
+        NOTE only valid for datasets with regular update cadences.
+        See `irregular_update_cadence` attribute for handling irregular cadence.
+
+        Returns
+        -------
+        np.timedelta64
```
ohsqueezy, line 675: We should drop the type since the type is annotated. See function_with_pep484_type_annotations at https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

gridded_etl_tools/utils/store.py
```diff
@@ -167,7 +167,11 @@ def mapper(self, refresh: bool = False, **kwargs: dict) -> fsspec.mapping.FSMap:
             A `MutableMapping` which is the S3 key/value store
         """
         if refresh or not hasattr(self, "_mapper"):
-            self._mapper = s3fs.S3Map(root=self.path, s3=self.fs())
+            if self.dm.custom_output_path:
```
ohsqueezy, line 170: This behavior at least needs to be documented in the docstring, since it will be confusing that `S3.path` returns a different path than the one used, and we should consider the overall design. I like that it uses an existing flag, and I understand that since it's an output path, it might make sense that the input path would be different, but that's not something we've done before on S3 and we don't have good documentation explaining it yet.

gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -30,15 +29,15 @@
...
+class Transform(Convenience):
     """
     Base class for transforming a collection of downloaded input files in NetCDF4 Classic format into
     (sequentially) kerchunk JSONs, a MultiZarr Kerchunk JSON, and finally an Xarray Dataset based on that MultiZarr.
     """
 
     # KERCHUNKING
 
-    def create_zarr_json(self, force_overwrite: bool = True):
+    def create_zarr_json(self, force_overwrite: bool = True, file_filter: str = None, outfile_path: str = None):
```
ohsqueezy, line 40: type annotation should be `Optional[str]`

gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -148,6 +135,72 @@ def kerchunkify(self, file_path: str, scan_indices: int = 0, local_file_path: Op
 
         return scanned_zarr_json
 
+    def local_kerchunk(self, file_path: str, scan_indices: int = 0):
```
ohsqueezy, line 138: The function definition is missing the annotation for the return value, and the annotation for scan_indices I think should be `int | slice` or `int | Tuple[int]`. There's a bit of info [here](https://stackoverflow.com/questions/76125621/how-to-annotate-slice-type-in-python) on why slice is hard to annotate. Also just a reminder we should be consistent about leaving the annotations out of the docstring.

gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -289,17 +323,25 @@ def parallel_subprocess_files(
         commands = []
         for existing_file in input_files:
             new_file = existing_file.with_suffix(replacement_suffix)
-            commands.append(  # map will convert the file names to strings because some command line tools (e.g. gdal) don't like Pathlib objects
-                    list(map(str, command_text + [existing_file, new_file]))
-                 )
+            if invert_file_order:
+                commands.append(  # map will convert the file names to strings because some command line tools (e.g. gdal) don't like Pathlib objects
+                        list(map(str, command_text + [new_file, existing_file]))
```
ohsqueezy, line 328: A bit minor, but `[new_file, existing_file]` should be stored in a variable to avoid copying the rest of the code.

gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -289,17 +323,25 @@ def parallel_subprocess_files(
...
         # Convert each comment to a Popen call b/c Popen doesn't block, hence processes will run in parallel
         # Only run 100 processes at a time to prevent BlockingIOErrors
         for index in range(0, len(commands), 100):
             commands_slice = [ Popen(cmd) for cmd in commands[index:index+100]]
             for command in commands_slice:
                 command.wait()
                 if not keep_originals:
-                    os.remove(command.args[-2])
+                    if not invert_file_order:
+                            os.remove(command.args[-2])
```
ohsqueezy, line 342: There's an extra tab indent in the formatting

gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -1009,14 +1065,54 @@ def is_append_contiguous(self, original_dataset: xr.Dataset, append_times: list[
...
+        since this is a regular pattern
+
+        Parameters
+        ----------
+        times
+            A datetime.datetime object representing the timestamp being checked
+
+        Returns
+        -------
+        bool | None
```
ohsqueezy, line 1087: I don't think this function could return `None` and it probably shouldn't anyway

gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -1009,14 +1065,54 @@ def is_append_contiguous(self, original_dataset: xr.Dataset, append_times: list[
...
+            Returns False for any unacceptable timestamp
+        """
+        previous_time = times[0]
+        for time in times[1:]:
+            if self.is_time_contiguous(time, previous_time) is not None:
```
ohsqueezy, line 1092: This seems confusing to me, I think is_time_contiguous should be edited to always return a bool. If the end of is_time_contiguous is reached, it should return True rather than the default return of None.

gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -1009,14 +1065,54 @@ def is_append_contiguous(self, original_dataset: xr.Dataset, append_times: list[
...
+        Returns
+        -------
+        bool | None
+            Returns False for any unacceptable timestamp
+        """
+        if self.irregular_update_cadence():
+            if not self.irregular_update_cadence()[0] <= (time - previous_time) <= self.irregular_update_cadence()[1]:
+                return False
+        elif time - previous_time != self.span_to_timedelta():
+            return False
```
ohsqueezy, line 1118: after this line is where it could return True

gridded_etl_tools/utils/store.py
rbanick: Do you think this switch should be inserted within S3’s `path` property instead? 

I think it’s slightly more useful to be able to manipulate both the default path and the custom path but ultimately I’m not opposed to moving it over there.

gridded_etl_tools/utils/zarr_methods.py
rbanick: I kind of prefer `str|None`. Cleaner looking and less to type. Thoughts?

gridded_etl_tools/utils/zarr_methods.py
rbanick: Putting slices in the docstrings caused all hell to break loose previously.

gridded_etl_tools/utils/zarr_methods.py
rbanick: If it returns true then `is_append_contiguous` will stop there instead of continuing onwards to the `are_times_contiguous` check. 

I think it’s better to just let it return None if nothing bad is encountered.

gridded_etl_tools/utils/zarr_methods.py
rbanick: I originally had the same plan but returning True makes it impossible to use `is_time_contiguous` as one step within larger functions as it short-circuits the enclosing function. See `is_append_contiguous`.

Try it out for yourself with the `are_times_contiguous` test.

gridded_etl_tools/utils/zarr_methods.py
ohsqueezy: Sure, `str | None` is good

gridded_etl_tools/utils/zarr_methods.py
ohsqueezy: ok

gridded_etl_tools/utils/store.py
ohsqueezy: Long term, I think we would want separate input and output path values and maybe we would want them to be arguments to the extract and parse functions or maybe even we would want them to be arguments to separate store objects which were arguments to the extract and parse functions.

For now, I think putting this in S3.path makes sense because that's where it is in Local.path. Is this being used to implement a custom S3 path or a custom local path? If we need to support custom S3 paths, I would put the code in S3.path and I would change the docstring in DM.__init__ to state that the parameter overrides both the local and S3 paths. If this is meant to be for a custom local path, however, I would just create a new Local store object instead and get the path through there.

rbanick: Sounds good. I’ve moved this logic w/ S3 and Local paths and amended the docstring.

gridded_etl_tools/dataset_manager.py
```diff
@@ -79,7 +79,7 @@ def __init__(
             Sets `DatasetManager.rebuild_requested`. If this parameter is set, the manager requests and parses all available data from beginning
             to end.
         custom_output_path : str, optional
-            Overrides the default path returned by `Convenience.output_path`
+            Overrides the default path returned by `Convenience.path` for Local and S3 stores.
```
ohsqueezy, line 82: The function name is still Convenience.output_path

rbanick: We’re actually overriding both. Correcting docstring.

gridded_etl_tools/utils/store.py
```diff
@@ -420,7 +419,10 @@ def path(self) -> pathlib.Path:
         pathlib.Path
             Path to the Zarr on the local filesystem
         """
-        return self.dm.output_path().joinpath(f"{self.dm.name()}.zarr")
+        if self.dm.custom_output_path:
+            return self.dm.custom_output_path
```
ohsqueezy, line 423: Currently, in Convenience.output_path, custom_path is not expected to contain a filename. It's just the directory where the data will be written, and the filename is generated by the class. That's probably how it should be treated by the store module as well, at least until we do a general clean up of input/output paths.

gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -1090,8 +1109,10 @@ def is_time_contiguous(self, time: datetime.datetime, previous_time: datetime.da
         """
         if self.irregular_update_cadence():
             if not self.irregular_update_cadence()[0] <= (time - previous_time) <= self.irregular_update_cadence()[1]:
+                self.info(f"Time value {time} and previous time {previous_time} are not consecutive")
```
ohsqueezy, line 1112: This probably should be self.warn instead of self.info

gridded_etl_tools/utils/store.py
rbanick: I don’t like this approach. During testing — the only time we normally use this flag — I find it useful to output files in the anticipated directory with different file names indicating key aspects of the test. For S3 in particular it would be unhelpful to have to create a new bucket/directory for each output.

ohsqueezy: That's fair, but it would be bad for two functions to be expecting different things from the same parameter. I took a quick look at tools and managers and I'm not sure anything is using Convenience.output_path besides the convenience module itself and the one line in this store module, so maybe we could just erase Convenience.output_path and define custom_output_path to be a path to a Zarr.

rbanick: The only place I can see using it is `create_output_path` so I’ll rewrite that.

gridded_etl_tools/dataset_manager.py
ohsqueezy: If you mean the function in store it's technically StoreInterface.path not Convenience.path

gridded_etl_tools/utils/store.py
ohsqueezy: I think that function can just be erased too

rbanick: Well we need it to automatically create the output path for Zarrs on the local file system, no?

ohsqueezy: I'm just going by looking at a grep of tools and managers code and I don't see anything using create_output_path

But also I just realized a Zarr is a directory so maybe it doesn't even matter

ohsqueezy: Might be worth cleaning out those functions though

rbanick: Wild. I’m down to drop `create_output_path` if it’s not being used.

ohsqueezy: writing to zarr uses the store module now for all stores so those convenience functions may not be used anymore

eschechter (approved):

PR #86, opened 2023-10-31
test: tests for dataset_manager

chrisrossi:
Get test coverage for dataset_manager up to 100%. This started with some lint, formatting, and continuous integration setup. To get coverage up to 100% in most cases tests were added, in some cases unused code was removed.


rbanick (dismissed):

tests/unit/conftest.py
```diff
@@ -0,0 +1,248 @@
...
+        "2022-01-28",
+        "2022-01-29",
+        "2022-01-30",
+        "2022-01-31",
+    ],
+    dtype="datetime64[ns]",
+)
+
+
+complex_update_times = np.array(
```
rbanick, line 265: This could be written as a mix of ranges and individual dates to make it clearer where the breaks and irregularity lie. But I’m not fussed if we want to do that another time.

tests/unit/test_dataset_manager.py
```diff
@@ -153,3 +171,31 @@ def test_transform_skip_prepare_input_files(manager_class):
...
+        assert John.get_subclass("George") is George
+        assert John.get_subclass("Ringo") is Ringo
+
+        assert Paul.get_subclass("George") is George
+        assert Paul.get_subclass("Ringo") is Ringo
+
+        assert George.get_subclass("Ringo") is Ringo
+
+    @staticmethod
+    def test_get_subclass_not_found():
```
rbanick, line 199: perhaps for good measure assert that `manager_class.get_subclass(“Pete”)` raises an error

rbanick (approved):

tests/unit/test_dataset_manager.py
chrisrossi: Well, returns None and raises a warning, but sure. 32a76ba

PR #88, opened 2023-11-03
revamp of contiguity checks

rbanick:


gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -1039,13 +1034,33 @@ def calculate_update_time_ranges(
...
+        """
+        Master function containing quality checks general to all datasets
+        If successful passes without comment. If unsuccessful raises a descriptive error message.
+
+        Parameters
+        ----------
+        original_dataset : xr.Dataset
+            The final dataset to be parsed
+        """
+        # Aggressively assert that the data is in contiguous time order. Only valid if the dataset's time dimension is longer than 1 element
```
ohsqueezy: We need to update this to explain that the times are checked to be in an order defined by the set

gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -1039,13 +1034,33 @@ def calculate_update_time_ranges(
...
+        # This is to protect against data corruption, especially during insert and append operations, but it should probably be replaced with 
+        # a more sophisticated set of flags that let the user decide how to handle time data at their own risk.
+        times = dataset[self.time_dim].values
+        if len(times) >= 2:
+            # Check is only valid if we have 2 or more values with which to calculate the delta
+            expected_delta = times[1] - times[0]
+            if not self.are_times_in_expected_order(times=times, expected_delta=expected_delta):
```
ohsqueezy: I think we should be branching here rather than passing the unused expected_delta value

gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -1067,74 +1074,37 @@ def update_quality_check(self,
...
+        expected_delta
+            Amount of time expected to be between each time
 
         Returns
         -------
         bool
-            Returns False for any unacceptable timestamp, otherwise True
+            Returns False for any unacceptable timestamp order, otherwise True
         """
+        # Warn if not using expected delta
```
ohsqueezy, line 1094: Related to the prev comment, I think splitting this into two functions would be better than printing a warning, maybe are_times_contiguous and are_times_in_cadence_order

ohsqueezy (approved):

PR #89, opened 2023-11-03
Dry run

rbanick:


gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -570,6 +570,12 @@ def to_zarr(self, dataset: xr.Dataset, *args, **kwargs):
             if not self.are_times_in_expected_order(times=times, expected_delta=expected_delta):
                 raise ValueError("Dataset does not contain contiguous time data")
 
+        # Exit script if dry_run specified
+        if self.dry_run:
+            self.info("Exiting without parsing since the dataset manager was instantiated as a dry run")
+            self.info(f"Dataset final state pre-parse:\n{dataset}")
+            return False
```
ohsqueezy, line 577: This is good but we want a different way of exiting instead of returning False since this function doesn't expect a return value, so I think a simple if/else with just the message in the if and the rest of the function in the else would be good

ohsqueezy (approved):

PR #91, opened 2023-11-08
test: test coverage for utils.attributes

chrisrossi:
This initial commit is primarily to show a strategy for converting abstract class attributes accessed by method calls to use simple attribute access instead. It turned out to be a little more involved than I'd hoped, mostly due to the requirement of preserving backwards compatibility. ("We do these things, not because they are easy, but because we thought they would be easy.")

This commit uses the ``name`` attribute as the first example for a working strategy going forward.

The primary innovation here is the introduction of the ``abstract_class_property`` class that is used to define abstract class properties. This is implemented independently of the ``abc`` package in the standard library, since it has no support for abstract class attributes.  One thing that's a little weird is figuring out when to check for concrete values for abstract attributes. With instance methods and attributes, the ``abc`` package just waits until instantiation to check the entire class to make sure it's concrete before instantiating. We do that, too, for these class properties, but since a class property can be accessed independtly of an instance, we also check individual properties as they are accessed. Some mechanisms for checking subclasses as they are declared were tried and discarded, as there's no real good way to know when the user has finished subclassing an abstract base class, unless they try to instantiate it.

As far as backwards compatibility, the first difficulty is we can't simply convert calls to ``DatasetManager.name()`` to ``DatasetManager.name``, since that would break downstream code that still uses the method call syntax. Unfortunately, that means that for each attribute we want to convert we need to introduce a new name. In this case ``name`` has been migrated to ``dataset_name``. Code which uses the ``name`` class method will generate a DeprecationWarning. Generally speaking, we should try to get rid of those warnings as they crop up in downstream code, by migrating that code sooner than later. When defining the ``dataset_name`` abstract class property, ``"name"`` is passed for the ``fallback`` argument. When the ``dataset_name`` property is accessed, first it will look for a concrete definition for ``dataset_name`` in the bound class. If it fails to find that, then it will fall back to looking for a ``name`` class method and use that (with a DeprecationWarning) if it finds one.

There is also a, seemingly circular, definition of the ``name`` classmethod in the ``Attributes`` mixin which uses the ``_find_fallback`` method to look for a ``dataset_name`` value. This is not for backwards compatibility with downstream code, as dataset managers that use ``name`` will also have defined ``name``, so this won't be used for those. This is here because ``gridded_etl_tools`` doesn't yet have 100% test coverage, so there is always some risk of some untested code in ``gridded_etl_tools`` using an older style class method instead of the property which has replaced it. When test coverage is up to 100%, it should be possible to remove ``_find_fallback`` and the fallback class methods that use it.

### DIFF
--- a/gridded_etl_tools/utils/attributes.py
+++ b/gridded_etl_tools/utils/attributes.py
@@ -1,6 +1,30 @@
-import numpy as np
 from abc import ABC, abstractmethod
+import warnings
+
+import deprecation
+import numpy as np
+
 from .store import StoreInterface, Local
+
+_NO_FALLBACK = object()
+
+
+class abstract_class_property(property):
+    def __init__(self, fallback=None):
+        self.fallback = fallback
+
+    def __get__(self, obj, cls):
+        if self.fallback is not None:
+            fallback = getattr(cls, self.fallback, None)
+            if fallback is not None:
+                warnings.warn(
+                    f"{cls.__name__}.{self.fallback}() is deprecated. Use {cls.__name__}.{self.name}.",
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
+                return fallback()
+
+        raise TypeError(f"No value in {cls.__name__} for abstract class attribute {self.name}")
 
 
 class Attributes(ABC):
@@ -9,6 +33,39 @@ class Attributes(ABC):
     These can be overriden in the ETL managers for a given ETL as needed
     """
 
+    def __init_subclass__(cls, **kwargs):
+        """Setup for abstract class properties."""
+        super().__init_subclass__(**kwargs)
+        for name, obj in list(cls.__dict__.items()):
+            if isinstance(obj, abstract_class_property):
+                # Tell property its own name
+                obj.name = name
+
+    @classmethod
+    def _check_abstract_class_properties(cls):
+        """Check that a subclass has provided concrete values for all required abstract class properties."""
+        for superclass in cls.mro():
+            for name, obj in superclass.__dict__.items():
+                if isinstance(obj, abstract_class_property):
+                    getattr(cls, name)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._check_abstract_class_properties()
+
+    @classmethod
+    def _find_fallback(cls, attr):
+        # There would have to be a class with an empty mro for branch coverage to be happy
+        for superclass in cls.mro():  # pragma NO BRANCH
+            value = superclass.__dict__.get(attr, _NO_FALLBACK)
+            if value is not _NO_FALLBACK:
+                if isinstance(value, abstract_class_property):
+                    break
+
+                return value
+
+        raise TypeError(f"No value in {cls.__name__} for abstract class attribute {attr}")
+
     @classmethod
     def host_organization(self) -> str:
         """
@@ -16,18 +73,15 @@ class Attributes(ABC):
         """
         return ""  # e.g. "Arbol"
 
+    dataset_name = abstract_class_property(fallback="name")
+    """
+    The name of each ETL is built recursively by appending each child class name to the inherited name
+    """
+
     @classmethod
-    @abstractmethod
+    @deprecation.deprecated("Use the dataset_name attribute")
     def name(cls) -> str:
-        """
-        The name of each ETL is built recursively by appending each child class name to the inherited name
-
-        Returns
-        -------
-        str
-           Name of dataset
-
-        """
+        return cls._find_fallback("dataset_name")
 
     @classmethod
     @abstractmethod
@@ -240,3 +294,7 @@ class Attributes(ABC):
             self._store = value
         else:
             raise ValueError(f"Store must be an instance of {StoreInterface}")
+
+
+# Won't get called automatically, because Attributes isn't a subclass of itself
+Attributes.__init_subclass__()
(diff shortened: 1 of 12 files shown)


gridded_etl_tools/utils/attributes.py
```diff
@@ -83,217 +84,148 @@ def host_organization(self) -> str:
...
-            The number of decimal places to round bounding box values to.
-        """
-        return 5
+    ensemble_numbers: list[int] = []
+    """
+    Numbers uses for ensemble, if any.
+    """
 
-    @property
-    def store(self) -> StoreInterface:
```
chrisrossi, line 208: Note that I removed the getter/setter for store because that is already handled well in the constructor for DatastoreManager.

ohsqueezy: That's true but at least with the current design or lack of design the store can be changed during the lifetime of the object

rbanick:
My comments follow @ohsqueezy ‘s — the work is impressive and it seems an elegant solution, but I do worry about the maintenance overhead.

gridded_etl_tools/utils/attributes.py
chrisrossi: Is that a common use case?

chrisrossi:
I think, maintenance-wise, it's fairly easy. We now just have a way to make abstract class properties as single-liners, the bit that does it is tested and works and likely doesn't need to be touched again. The ugliest bits are the backwards compatibility shims that can be removed over time.

gridded_etl_tools/utils/attributes.py
ohsqueezy: It's not and maybe even shouldn't be but in general the library runs in the interpreter and notebooks more often than originally, so we've been tightening up the member properties like keeping the store from being set to something arbitrary

chrisrossi: e99bba7

ohsqueezy: :pray:

ohsqueezy:
Along the lines of the discussion about the store property, we've noted a couple of times where the fact that you can't set a @property to another value without a setter has actually been a good thing, making these properties somewhat like constants, but if these properties are changed to regular class variables, they will be open to being changed.

ohsqueezy:
I ran the gridded-etl-managers tests which are mostly systems tests and they passed against this branch, so it seems like it's at least an option to drop `@property` if we want. I think either way, testing and deprecation will be very beneficial.

chrisrossi:
> Along the lines of the discussion about the store property, we've noted a couple of times where the fact that you can't set a `@property` to another value without a setter has actually been a good thing, making these properties somewhat like constants, but if these properties are changed to regular class variables, they will be open to being changed.

True. I do like a good guard rail and keeping users from accidentally breaking things is an admirable goal. I don't think these particular attributes are particularly high risk or high stakes, so it's probably ok to trust the user on these. The balancing act here, o of course, is finding the right balance of protecting users from themselves versus keeping the code simple and readable.

If we like the read-only property of properties with no setters, we can also use a simple descriptor for that purpose. In commit 46e1909, there's a ``readonly_property`` descriptor that allows us to keep attribute declarations to a single line and keep the code readable, while preventing accidental runtime mutation. It has only been applied to one attribute in this commit, but if we like the pattern it can be extended. I don't personally think it's necessary, but it's also succinct enough that I don't mind having it in there if it imparts some extra peace of mind.

chrisrossi:
> I ran the gridded-etl-managers tests which are mostly systems tests and they passed against this branch, so it seems like it's at least an option to drop `@property` if we want. I think either way, testing and deprecation will be very beneficial.

Excellent! Thanks for checking that.

chrisrossi:
> 
> If we like the read-only property of properties with no setters, we can also use a simple descriptor for that purpose. In commit [46e1909](https://github.com/Arbol-Project/gridded-etl-tools/commit/46e1909c6474e35f7c0a5d00f8feed88bef14652), there's a `readonly_property` descriptor ...

Following up on this, it turns out it doesn't really work for classes, just instances, so I'm likely to back this one out. 

What this means is that:

```
dm = MyManagerClass()
dm.organization = "Hydra"  # <-- will raise an AttributeError

MyManagerClass.organization = "Hydra"   # <-- Doesn't raise anything
dm = MyManagerClass()
dm.organization == "Hydra"   # <-- True


```

ohsqueezy:
Taking a closer look at the attributes, I think maybe all of them besides `store` should all be class properties anyway. We can ask @rbanick to confirm. For example, `spatial_precision` is always the same for every instance of its specific dataset class.

rbanick:
@ohsqueezy is right, we almost never change the class properties in production. Occasionally I’ll tweak them during development and debugging — for instance reducing forecast_hours to just a few forecasts — but that can be replicated w/in the manager if needed.

rbanick:
@chrisrossi we’ll need to make a downstream PR aligning all the managers w/ the new system before we can merge this. 

@ohsqueezy do you think you can take charge of that or should I do it?

ohsqueezy:
Because of the deprecation we probably don't have to make any changes to managers. We can just update managers as we go and make sure to adopt these attributes with new ETLs. I just ran the managers test using this branch and they passed (we're working on automating this test in the future if we can sort out the privacy details). But I think @rbanick should merge in case it does end up breaking anything in production.

chrisrossi:
Yep, that's why I jumped through all those hoops to put in backwards
compatibility shims.

rbanick:
ah got it got it. OK I’m setting up a staging environment now, let me test a few dry runs in that today/tomorrow and I’ll merge if all is well.

PR #93, opened 2023-11-13
Quality checks

rbanick:
Formalizing pre-parse quality checks for datasets. Checking:
* NaN Values
* Extreme Values
* Misspecified dataset dtype
* Time value contiguity

Note these must pass extensive system tests for downstream ETL managers to ensure they work with production ETLs


examples/managers/my_new_etl.py
```diff
@@ -330,6 +330,7 @@ def name(cls) -> str:
     def relative_path(self) -> pathlib.Path:
         return super().relative_path() / "min"
 
+    @classmethod
     def data_var(self) -> str:
```
ohsqueezy, line 334: For class methods by convention use `cls` instead of `self` just because the first arg will be a class not an instance. We might also want to adopt `abstract_class_property` from https://github.com/Arbol-Project/gridded-etl-tools/pull/91

gridded_etl_tools/utils/attributes.py
```diff
@@ -182,6 +187,13 @@ def hindcast_steps(self) -> list[int]:
         """To be overwritten by actual hindcast datasets"""
         return list(None)
 
+    @property
```
ohsqueezy, line 190: I think every dataset class will want to always have this same property, so we would want this to be a class property. Related to this, I posted a comment that maybe all of the attributes besides store should be class properties https://github.com/Arbol-Project/gridded-etl-tools/pull/91#issuecomment-1808791840

gridded_etl_tools/utils/convenience.py
```diff
@@ -632,4 +635,31 @@ def span_to_timedelta(cls) -> np.timedelta64:
...
+        return span_to_td
+
+    def get_random_coords(self, dataset: xr.Dataset) -> dict[str, Any]:
+        """
+        Derive a dictionary of random coordinates, one for each coordinate in the input dataset
+
+        Parameters
+        ----------
+        dataset
+            An Xarray dataset
```
ohsqueezy, line 647: don't forget the return val in docstring

gridded_etl_tools/utils/convenience.py
```diff
@@ -632,4 +635,31 @@ def span_to_timedelta(cls) -> np.timedelta64:
...
+        """
+        coords_dict = {}
+        # We select from dims, not coords because inserts drop all non-time_dim coords
+        for dim in dataset.dims:
+            coords_dict.update({dim : random.choice(dataset[dim].values)})
+        return coords_dict
+    
+    @property
+    def extreme_values_by_unit(self):
```
ohsqueezy, line 656: If these are intended to be the same for all datasets (basically a global var) they probably belong defined at the module level, then accessed by `convenience.extreme_values_by_unit` rather than `convenience.Convenience.extreme_values_by_unit`/`self.extreme_values_by_unit`

gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -1126,3 +1169,10 @@ def are_times_in_expected_order(self, times: tuple[datetime.datetime], expected_
             previous_time = instant
         # Return True if no problems found
         return True
+
+    def post_parse_quality_check(self):
```
ohsqueezy, line 1173: I think we want to this to run in Prefect where the possibility of an auto revert or a change to API retrieval could happen. Currently this repo doesn't have any knowledge of Prefect or the API and I think we want to keep it that way.

ohsqueezy (commented):
Mainly I think the post check belongs in Prefect, I added a comment about that. Otherwise just minor comments.

gridded_etl_tools/utils/attributes.py
rbanick: By a class property you mean use the system Chris has set up in #91 ? Makes sense to me but let’s get this PR merged and then I’ll work on integrating that PR into our managers and attributes.

gridded_etl_tools/utils/zarr_methods.py
rbanick: Yeah agreed the more I thought about this the more it belongs as a standalone script we run in Prefect. I’ll remove this function and we can implement it elsewhere

gridded_etl_tools/utils/convenience.py
rbanick: I’m hesitant to introduce a new pattern like this into the `tools` repo. The convenience of calling every property with `self` is a major plus of the current setup, even when those properties are in effect static.

ohsqueezy: It's just a bit different than the other properties because it's not meant to be override-able per dataset, but I guess it's fine if it lives here

gridded_etl_tools/utils/convenience.py
```diff
@@ -632,4 +635,41 @@ def span_to_timedelta(cls) -> np.timedelta64:
             cls.SPAN_MONTHLY: np.timedelta64(1, "M"),
             cls.SPAN_YEARLY: np.timedelta64(1, "Y"),
         }
-        return span_to_td[cls.temporal_resolution()]
```
ohsqueezy, line 635: Was this meant to be changed

rbanick: yes because we realized basing this on the temporal resolution didn’t always make sense.

In reality we are no longer using this class method in live code but it seems useful enough to warrant preservation for the time being.

rbanick: Yep will do

PR #95, opened 2023-11-16
Post parse qc

rbanick:


gridded_etl_tools/dataset_manager.py
```diff
@@ -120,9 +121,12 @@ def __init__(
             will fail with a warning message.
         use_local_zarr_jsons: bool, optional
             Write out Zarr JSONs created via Kerchunk to the local file system. For use with remotely kerchunked
-            datasets.
+            datasets. Defaults to True to speed up post-parse quality checks.
```
ohsqueezy, line 124: This param actually defaults to False which also does seem like the reasonable default, but if it makes more sense what it could do is default to None and then unless it is manually set by the caller, it gets set to True when skip_post_parse_qc is False or False otherwise

gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -1209,3 +1212,172 @@ def are_times_in_expected_order(self, times: tuple[datetime.datetime], expected_
...
+        Parameters
+        ----------
+        checks
+            The number of values to check. Defaults to 100.
+        threshold
+            The tolerance for diversions between original and parsed values.
+            Absolute differences between them beyond this limit will raise a ValueError
+        """
+        if self.skip_post_parse_qc:
+            self.info("Skipping post-parse quality check since the source data is different from the parsed data")
```
ohsqueezy, line 1229: The caller can set skip_post_parse_qc to True for any reason though, so I think either change that behavior to not being configurable or this message should just say something generic like Skipping post-parse QC because skip_post_parse_qc is set

gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -1209,3 +1212,172 @@ def are_times_in_expected_order(self, times: tuple[datetime.datetime], expected_
...
+            The original dataset, reformatted similarly to the production dataset
+        """
+        # Drop unused coords and dims
+        for coord in ["crs"]:
+            if coord in orig_ds:
+                orig_ds = orig_ds.drop(coord)
+        # Rename coordinates
+        orig_data_var = [var for var in orig_ds.data_vars][0]
+        rename_coords = {orig_data_var: self.data_var()}
+        for coord in ["lat", "lon"]:
```
ohsqueezy, line 1321: If I'm understanding correctly, I'm a little concerned about this being hard-coded to only support the literal names lat/lon and latitude/longitude. If an ETL doesn't fit that paradigm it either has to completely rewrite this function to override it or it has to force skip the QC. For example, I have seen "long" in the wild. I think maybe the first half of this function is something to be standardized as standarize_coords or something like that, and then if we need to change the literals we could do it for every set there.

gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -1209,3 +1212,172 @@ def are_times_in_expected_order(self, times: tuple[datetime.datetime], expected_
...
+        ----------
+        orig_ds
+            The original dataset, unformatted
+        orig_file_path
+            The pathlib.Path to the randomly selected original file
+        """
+        orig_file_path = random.choice(list(self.input_files()))
+        if self.protocol == "file":
+            raw_ds = xr.open_dataset(orig_file_path)
+        # Presumes that use_local_zarr_jsons is enabled. This avoids repeating the DL from S#
```
ohsqueezy, line 1290: We don't necessarily have to assume, if it makes sense we could just skip or throw an exception depending on the importance if use_local_zarr_jsons is not enabled

ohsqueezy (approved):
I'm OK with approving especially with the through testing, but I wanted to share a few thoughts that don't necessarily need to be addressed specifically in this PR.

One thing I'm concerned about is that the checks are not generic enough to apply to all data and either we'll start seeing ETLs crash in the post-check that are actually fine (though we can address those cases as they happen) or possibly worse we'll start taking the same approach we did with "verify" where we just set ETLs to skip the check. 

More generically, I don't know that we have sufficient enforcement that the input data conforms to things like it has to be open-able in Xarray or it uses one of either lat/lon or latitude/longitude. I guess these theoretical sets could just skip QC, but I'm wondering if maybe one of these ideas make sense 1) enforce some characteristics of the input data that distinguish what this library supports or 2) have a Checkable class/trait that sets can choose to either belong to or not

gridded_etl_tools/utils/zarr_methods.py
rbanick: I’m into this

gridded_etl_tools/dataset_manager.py
rbanick: It’s actually not a good idea to set it to

gridded_etl_tools/utils/zarr_methods.py
rbanick: That makes sense to me.

An alternative approach would be to call `postprocess_zarr` on every dataset and see what happens. I think at least 95% of this step is managed by that single function. I’m just not completely sure it covers every eventuality. Let me test.

PR #97, opened 2023-11-21
fix: consolidate package setup to pyproject.toml

chrisrossi:
This was always a good idea, but with the release of setuptools 69.0.1, it became a necessity, as the new version breaks with setup spread out over both setup.cfg and pyproject.toml.


noxfile.py
```diff
@@ -38,7 +38,7 @@ def cover(session):
 
 @nox.session(py=DEFAULT_INTERPRETER)
 def lint(session):
-    session.install("black", "flake8")
+    session.install("black", "flake8", "flake8-pyproject")
```
chrisrossi, line 41: Apparently the maintainers of flake8 don't believe in pyproject.toml, so this plugin is required for now.

chrisrossi:
This was primarily done by running [ini2toml](https://pypi.org/project/ini2toml/) on setup.cfg and merging the output into pyproject.toml by hand. It ended up moving the long description into a README section that setuptools didn't like at all, so I removed that and added 'readme' value to [project] that points to README.md. 

There was some initial confusion about what to do with the version, but it looks like with this configuration setuptools_scm is free to set it. It's possible I'm missing something here, though.

noxfile.py
rbanick: Wild?

rbanick (approved):

PR #105, opened 2023-11-30
proposed nox changes

rbanick:


noxfile.py
```diff
@@ -15,17 +15,17 @@
...
     session.run(
         "pytest",
         f"--cov={CODE}",
-        "--cov=tests.unit",
+        "--cov=gridded_etl_tools",
         "--cov-append",
         "--cov-config",
         HERE / ".coveragerc",
         "--cov-report=term-missing",
-        "tests/unit",
```
chrisrossi, line 28: The system tests aren't included in coverage, so shouldn't be included.

noxfile.py
```diff
@@ -15,17 +15,17 @@
...
+        "tests",
```
chrisrossi, line 28: The unit tests are distinct from the system tests. We don't use them to measure coverage and the system tests are expected to take longer to run and require extra setup beyond what can be scripted in Nox, so we make them optional.

PR #110, opened 2023-12-07
Write .zattrs and .zmetadata JSON directly to indicate update_in_progress

ohsqueezy:
This removes the empty_dataset metadata-only write hack in favor of directly writing the .zattrs and .zmetadata JSON to do a metadata-only write.

This is working in tests with S3 and local. Before to_zarr, update_in_progress is True in the dataset attrs in xarray (checked by calling DM.store.dataset), and after to_zarr update_in_progress is False.

### DIFF
--- a/gridded_etl_tools/utils/store.py
+++ b/gridded_etl_tools/utils/store.py
@@ -85,6 +85,30 @@ class StoreInterface(ABC):
         else:
             return None
 
+    @abstractmethod
+    def write_metadata_only(self, attributes: dict):
+        """
+        Writes the metadata to the stored Zarr. Not available to datasets on the IPLD store. Use DatasetManager.to_zarr
+        instead.
+
+        Open the Zarr's `.zmetadata` and `.zattr` files with the JSON library, update the values with the values in the
+        given dict, and write the files.
+
+        These changes will be reflected in the attributes dict of subsequent calls to `DatasetManager.store.dataset`
+        without needing to call `DatasetManager.to_zarr`.
+
+        Parameters
+        ----------
+        attributes
+            A dict of metadata attributes to add or update to the Zarr
+
+        Raises
+        ------
+        NotImplementedError
+            If the store is IPLD
+        """
+        pass
+
 
 class S3(StoreInterface):
     """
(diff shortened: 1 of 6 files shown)


gridded_etl_tools/utils/store.py
```diff
@@ -280,6 +305,25 @@ def get_metadata_path(self, title: str, stac_type: str) -> str:
...
+
+            # Update given attributes at the appropriate location depending on which z file
+            if z_path == ".zmetadata":
+                current_attributes["metadata"][".zattrs"].update(update_attrs)
+            else:
+                current_attributes.update(update_attrs)
+
+            # Write back to Zarr
+            with self.fs().open(f"{self.path}/{z_path}", "w") as z_contents:
+                json.dump(current_attributes, z_contents)
```
eschechter, line 325: There's no need to instantiate a blank dict at the start of the for loop. Just set `current_attributes = json.load(z_contents)` on line 315,

gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -571,32 +571,56 @@ def to_zarr(self, dataset: xr.Dataset, *args, **kwargs):
...
-            if not isinstance(self.store, IPLD):
+            if not isinstance(self.store, IPLD) and exists_at_start:
                 # Indicate in metadata that update is complete.
-                empty_dataset.attrs["update_in_progress"] = False
-                self.info("Re-writing Zarr to indicate in the metadata that update is no longer in process.")
-                empty_dataset.to_zarr(self.store.mapper(), append_dim=self.time_dim)
+                self.info("Writing metadata after writing data to indicate write is finished.")
+                self.store.write_metadata_only(update_attrs=post_parse_attrs)
+
+    def post_parse_attrs(self, dataset: xr.Dataset = None) -> dict[str, Any]:
```
eschechter, line 600: 1. I think this should be rewritten to be more functional instead of relying on a [side effect](https://en.wikipedia.org/wiki/Side_effect_(computer_science)). In other words, it should return the dataset with the attrs removed as opposed to just removing them. Then the function name could be something like `move_post_parse_attrs_to_dict`.
2. I don't think `dataset` should be an optional key word arg, I think it's required.

gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -620,7 +620,7 @@ def post_parse_attrs(self, dataset: xr.Dataset = None) -> dict[str, Any]:
                 # For example "date range" should only be updated after a successful parse
                 update_attrs[attr] = dataset.attrs.pop(attr, None)
 
-        return update_attrs
+        return dataset, update_attrs
```
ohsqueezy, line 623: This is still modifying the input dataset during the attrs.pop line. I think if the dataset is copied first using a shallow copy with xarray.Dataset.copy you could run pop without modifying the input.

gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -612,6 +612,7 @@ def move_post_parse_attrs_to_dict(self, dataset: xr.Dataset) -> dict[str, Any]:
         update_attrs
```
ohsqueezy, line 612: minor thing but the typing and docstring need updating

eschechter (approved):

PR #115, opened 2023-12-19
pinning new ipldstore version

rbanick:


eschechter (approved):

chrisrossi:
Something to think about for the future, is maybe we could start releasing ipldstore to pypi (or create our own repo). The advantage of that is we can avoid pinning specific versions, and just use something like "ipldsore >= 2.1.2". 

Generally speaking, you want to use as permissive a version spec as possible in any dependencies. The more restrictive the version specifiers are, the more likely you are to run into issues later with a dependency needing a different version of the library than your project does and causing the dependency resolver to puke.

rbanick:
@chrisrossi agreed and it’s something we’ve thought about. Since much of the original ipldstore was forked from an outside dev’s github project we didn’t feel comfortable releasing as our project, even after making substantial modifications for our purposes. But it’s worth discussing now, almost a year down the road.

chrisrossi:
Another possibility that might avoid pain without having to release to pypi is to point to maintain a branch that points to the latest stable release, so you don't have to update the version in each downstream project.

Something like:
    "ipldstore @ git+https://github.com/dClimate/ipldstore@latest-stable",

PR #122, opened 2024-01-15
Consolidate extract functions into Extractor class

ohsqueezy:
Multithreaded extract code common to most managers has been generalized into the main Extractor class.

A child Extractor FTPExtractor generalizes a lot of the extract step for FTP ETLs.

This was [previously being worked on in gridded-managers](https://github.com/Arbol-Project/gridded-etl-managers/pull/135) and there are implementations for CPC, ERA5, and CHIRPS in that repository.

This is not meant to implement the alternative to mix-ins that was discussed at the 2024 planning meeting. It is work from 2023 that hopefully will be the last utility before we refactor to implement a more unified design.

### DIFF
--- a/gridded_etl_tools/utils/extractor.py
+++ b/gridded_etl_tools/utils/extractor.py
@@ -1,0 +1,363 @@
+"""
+Consolidation of functions useful during the extract step of an ETL cycle for a dataset manager.
+"""
+
+# The annotations dict and TYPE_CHECKING var are necessary for referencing types that aren't fully imported yet. See
+# https://peps.python.org/pep-0563/
+from __future__ import annotations
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from .. import dataset_manager
+
+import pathlib
+import typing
+import ftplib
+import re
+import time
+import multiprocessing
+
+
+class Extractor:
+    """
+    Base class for common extraction functions. Associate this class with a `DatasetManager` object by passing the
+    object at initialization.
+
+    Extractor's functions are used to request source data, using multiprocessing when possible. `Extractor.pool` is
+    used to launch the requests using a thread pool.
+
+    If more specific types of extraction are available in a child class, consider using the child class instead. For
+    example, if extracting from FTP, use FTPExtractor.
+    """
+
+    def __init__(self, dm: dataset_manager.DatasetManager):
+        """
+        Create a new Extractor object by associating a Dataset Manager with it.
+
+        Parameters
+        ----------
+        dm
+            Source data for this dataset manager will be extracted
+        """
+        self.dm = dm
+
+    def pool(self, batch_processor: typing.Callable[..., bool], batch: typing.Sequence[typing.Sequence]) -> bool:
+        """
+        Launch a batch of requests simultaneously, wait for them all to complete, then return a boolean indicating
+        whether any of the requests were successful.
+
+        The batch processor is any generic callable that accepts the arguments provided in each entry of the batch.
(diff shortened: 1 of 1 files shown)


gridded_etl_tools/utils/extractor.py
```diff
@@ -0,0 +1,363 @@
...
+        """
+        When called like a function, set the host parameter. See `FTPExtractor.__enter__`.
+
+        Parameters
+        ----------
+        host
+            Address to connect to for source data
+        """
+        self.host = host
+        return self
```
eschechter, line 237: I think the more readable way to do this is to have `host` be a parameter to `__enter__` and set the instance variable there

gridded_etl_tools/utils/extractor.py
```diff
@@ -0,0 +1,363 @@
...
+        pathlib.Path
+            The object's working directory on the FTP server
+
+        Raises
+        ------
+        RuntimeError
+            If the FTP connection is not open yet
+        """
+        try:
+            return pathlib.Path(self.ftp.pwd())
```
eschechter, line 253: I don't think we should use `pathlib.Path`s to represent the paths on the FTP server. If someone were to run this code on windows, `str(path)` would concatenate the parts of the path with backslashes, resulting in an invalid FTP path. I would use `pathlib.PurePosixPath`

gridded_etl_tools/utils/extractor.py
```diff
@@ -0,0 +1,363 @@
...
+            the connection being closed.
+        """
+        if not self.ftp.nlst(str(path)):
+            raise RuntimeError(f'Could not find path "{path}" on FTP server.')
+        try:
+            self.ftp.cwd(str(path))
+        except ftplib.error_perm:
+            raise RuntimeError("Error changing directory. Is the FTP connection open?")
+
+    def request(self, source: pathlib.Path, destination: pathlib.Path = pathlib.Path()) -> bool:
```
eschechter, line 283: Would it be worth it to log a warning saying that this won't actually result in parallel downloading of files? I could see a user being confused as to why the FTPExtractor is downloading one file at a time

gridded_etl_tools/utils/extractor.py
ohsqueezy: Are you sure? That's what I tried originally, and I couldn't pass an argument. I ended up using the method [here](https://stackoverflow.com/a/68300883) which mentions `__enter__` doesn't accept arguments.

eschechter: My mistake - I was mixing up args to `__enter__` with the technique where you pass the args in `__init__` in a one liner with the `with` statement.

eschechter (approved):

gridded_etl_tools/utils/extractor.py
```diff
@@ -0,0 +1,373 @@
...
+        thread_count = max(1, multiprocessing.cpu_count() - 1)
+
+        # run downloads
+        if len(batch) > 0:
+            self.dm.info(f"Submitting request for {len(batch)} datasets in parallel using {thread_count} threads.")
+
+            # download in parallel
+            with multiprocessing.pool.ThreadPool(processes=thread_count) as pool:
+                for result in pool.starmap(batch_processor, batch):
+                    if result:
```
chrisrossi, line 85: If any download succeeds the operation is considered as having succeeded? The operation only fails if all downloads fail?

ohsqueezy:
Replaced by https://github.com/Arbol-Project/gridded-etl-tools/pull/127

PR #123, opened 2024-01-17
Reflect new extractor tool in CHIRPS example manager

rbanick:
Modifying the example CHIRPS manager to use the new extractor toolkit and an FTP interface


ohsqueezy:
This looks good. We can merge it to extractor and if it fails the tests, leave it for a separate PR.

PR #125, opened 2024-01-17
tests: tests for utils.zarr_methods part 5

chrisrossi:


tests/unit/utils/test_zarr_methods.py
```diff
@@ -1337,3 +1337,347 @@ def test_update_zarr(manager_class):
         dm.update_parse_operations.assert_called_once_with(
             original_dataset, update_dataset, insert_times, append_times
         )
+
+    @staticmethod
+    def test_update_setup(manager_class, fake_original_dataset, fake_complex_update_dataset):
+        dm = manager_class()
+        insert_times, update_times = dm.update_setup(fake_original_dataset, fake_complex_update_dataset)
+        assert insert_times == [
```
rbanick, line 1345: For ease of maintenance going forward could we represent the blocks of continuous dates as ranges, e.g. `numpy.arange(date, date1, timedelta64(1, “[D]”)` as you do below? And then merge them all? That’s much easier to read and understand at a glance then this big block of dates (which I built back in the day, doh).

rbanick: Since the point is to push the update mechanism to elegantly divide up contiguous ranges of dates it’s really important to quickly see those date blocks and outlier single dates.

tests/unit/utils/test_zarr_methods.py
```diff
@@ -1337,3 +1337,347 @@ def test_update_zarr(manager_class):
...
+        assert len(dataset.time) > len(time_values)
+
+        dataset = dm.prep_update_dataset(dataset, time_values, chunks)
+
+        assert numpy.array_equal(dataset.time, time_values)
+        dataset.chunks["time"][0] == 5
+        dataset.chunks["latitude"][0] == 4
+        dataset.chunks["longitude"][0] == 4
+
+        # TODO: I'd like to be able to verify order of coordinates after transpose, but can't figure out how get
```
rbanick, line 1624: Perhaps you could look at the order of chunks / coordinates for the data variable array? This is what really matters, after all, as coordinate order there affects writing and retrieval speeds.

rbanick (approved):

tests/unit/utils/test_zarr_methods.py
chrisrossi: I almost did that, but wasn't sure it would actually be more readable given the number of non-contiguous regions. It's still kinda ugly but maybe a little better?

7f81e87

rbanick: Yeah I think that’s better. There’s no world where this is super readable but at least you can identify the breaks easily.

PR #127, opened 2024-01-22
Extractor w/ tests

rbanick:
Cleaner branch of extractor tests


examples/managers/chirps.py
```diff
@@ -7,7 +7,7 @@
 import pathlib
 import xarray as xr
 from gridded_etl_tools.dataset_manager import DatasetManager
-from gridded_etl_tools.utils.extractor import FTPExtractor
+from gridded_etl_tools.utils.extractor import FTPExtractor, pool
```
ohsqueezy, line 10: Directly importing a name like pool can create name conflicts, I would consider calling pool as `extractor.pool` in CHIRPS and in the tests

```
import gridded_etl_tools.utils.extractor as extractor
```

gridded_etl_tools/utils/extractor.py
```diff
@@ -16,92 +16,86 @@
...
-    Base class for common extraction functions. Associate this class with a `DatasetManager` object by passing the
-    object at initialization.
+log = logging.getLogger("extraction_logs")
 
-    Extractor's functions are used to request source data, using multiprocessing when possible. `Extractor.pool` is
-    used to launch the requests using a thread pool.
 
-    If more specific types of extraction are available in a child class, consider using the child class instead. For
-    example, if extracting from FTP, use FTPExtractor.
+@staticmethod
```
ohsqueezy, line 25: The @staticmethod header is only relevant inside a class as far as I know

gridded_etl_tools/utils/extractor.py
```diff
@@ -16,92 +16,86 @@
...
-        Returns
-        -------
-        bool
-            `True` if any of the batch calls returns `True`, `False` otherwise
-        """
-        # Success remains false until the first successful request is completed
+        log.info(f"Downloading took {(time.time() - start_downloading) / 60:.2f} minutes")
+    else:
+        log.info("No requests prepared to be submitted, please check batch_processor is functioning properly.")
         success = False
```
ohsqueezy, line 73: I thought we were removing the return value

gridded_etl_tools/utils/extractor.py
```diff
@@ -187,6 +181,17 @@ class FTPExtractor(Extractor):
     # Used to limit retrieval to a single thread at a time
     semaphore: multiprocessing.synchronize.Semaphore = multiprocessing.Semaphore()
 
+    def __init__(self, host: str):
+        """
+        When called like a function, set the host parameter. See `FTPExtractor.__enter__`.
```
ohsqueezy, line 186: Description needs to be updated and I don't think __enter__ needs to be referenced anymore

ohsqueezy (approved):

gridded_etl_tools/utils/extractor.py
rbanick: I rethought this a bit as I was editing it. I don’t think returning `True` if there are *any* successes will ever be useful, but I think there are scenarios where returning False in the event there are *any* failures could be useful.

PR #128, opened 2024-01-22
Speed up pre parse

eschechter:


gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -861,6 +864,7 @@ def update_zarr(self):
         """
         original_dataset = self.store.dataset()
         update_dataset = self.transformed_dataset()
+        self.pre_chunk_dataset = update_dataset
```
rbanick, line 867: I think you need to create this before `update_dataset.chunk(new_chunks)` on line 1042 (of the current `zarr_methods` file) as well.

PR #130, opened 2024-01-23
Extractor full coverage

chrisrossi:
This was arrived at by running tests with coverage reporting over and over again, fixing each missing line or branch in turn, until 100% coverage was reached:

`$ pytest --cov gridded_etl_tools.utils.extractor tests/unit/utils/test_extractor.py`

Because the primary mission objective for this PR is pedagogy, every change get its own commit so you can follow the progression. Each commit corresponds to looking at the first line number in the missing coverage report and figuring out how to cover it.

### DIFF
--- a/tests/unit/utils/test_extractor.py
+++ b/tests/unit/utils/test_extractor.py
@@ -185,11 +185,28 @@ class TestFTPExtractor:
         host = "what a great host"
 
         with extract(host) as ftp:
+            assert ftp.cwd == pathlib.PosixPath("")
             ftp.cwd = "over there"
-            ftp.cwd
 
         ftp_client.pwd.assert_called_once()
         ftp_client.cwd.assert_called_once_with("over there")
+
+    @staticmethod
+    def test_cwd_setter_no_such_path(manager_class):
+        extract = extractor.FTPExtractor(manager_class())
+        ftp_client = ftplib.FTP = DummyFtpClient()
+        ftp_client.pwd = Mock(return_value="")
+        ftp_client.nlst = Mock(return_value=None)
+
+        host = "what a great host"
+
+        with extract(host) as ftp:
+            with pytest.raises(RuntimeError):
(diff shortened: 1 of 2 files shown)


tests/unit/utils/test_extractor.py
```diff
@@ -208,6 +208,22 @@ def test_cwd_setter_no_such_path(manager_class):
         ftp_client.cwd.assert_not_called()
         ftp_client.nlst.assert_called_once_with("over there")
 
+    @staticmethod
+    def test_cwd_client_error(manager_class):
+        extract = extractor.FTPExtractor(manager_class())
+        ftp_client = ftplib.FTP = DummyFtpClient()
+        ftp_client.pwd = Mock(return_value="")
+        ftp_client.cwd = Mock(side_effect=ftplib.error_perm)
```
rbanick, line 215: This doesn’t really cover the case where `ftp` is called outside of a context manager, or more specifically after a context manager has expired, which was what I was stuck on. Perhaps that’s not necessary to test but forcing the `ftplib.error_perm` error like this seems to skirt around the issue?

chrisrossi: This may be a case where not being familiar with the code gives me a bit of an advantage. Remember, a unit test tests only the unit under test--ie the code in the function being tested, not any of its dependencies. In this case, there is a branch where something happens when a particular type of exception is raised so I write a test where that exception gets raised. From the standpoint of unit testing, that's all we have to do. 

As a person unfamiliar with the code, I don't know why or under what circumstances that exception gets raised. More importantly, though, the code also doesn't know. There's just an `except` clause that gets executed when a particular thing happens. The code doesn't need to know why that happens, so neither does the test for that code. Thinking about it at the higher level of "oh, this is going to happen if they user tries to set `cwd` after closing the FTP connection" actually just distracts us from the main point--write a test that exercises those lines of code, then move on.

PR #131, opened 2024-01-23
Simplify parallel subprocessing w/ CDO, optimize pre_parse_quality_check, consolidate dataset_category attribute

rbanick:
* Pre-parse quality checks were unnecessarily chunking for every random value check. Use an instance variable of an unchunked dataset to avoid this significant performance slowdown
* The CDO command line tool is capable, with nudging, of converting directly to our preferred NetCDF4 Classic format directly from source files. This obviates the need for an additional `nc_to_nc4s` step in many ETLs with considerable cost savings.
* Consolidate various properties specifying a dataset’s category into a single `dataset_category` attribute.
* Fixing edge cases where `post_parse_quality_check` would issue false positives (approve mismatched data)


gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -427,8 +435,13 @@ def convert_to_lowest_common_time_denom(self, raw_files: list, keep_originals: b
...
-        self.parallel_subprocess_files(raw_files, command_text, "", keep_originals)
+        command_text = ["cdo", "-f", "nc4c", "splitsel,1"]
+        self.parallel_subprocess_files(
+            input_files=raw_files,
+            command_text=command_text,
+            replacement_suffix=".nc4",
+            keep_originals=keep_originals,
+        )
 
     def ncs_to_nc4s(self, keep_originals: bool = False):
```
eschechter, line 446: Maybe make note of the fact that `ncs_to_nc4s` is no longer in use

gridded_etl_tools/utils/attributes.py
```diff
@@ -219,6 +219,18 @@ def missing_value_indicator(cls) -> str:
     Tags for dataset.
     """
 
+    dataset_category: str = "observation"
```
eschechter, line 222: You should use `typing.Literal` instead of `str` for the type, since we want to clue the user into the fact that we want them to use a limited number of strings when overwriting

eschechter (approved):

gridded_etl_tools/utils/zarr_methods.py
rbanick: it’s actually still used by some managers that don’t convert via CDO

PR #137, opened 2024-02-01
tests: tests for utils.store

chrisrossi:
Probably the S3 and Local stores could be refactored to share most of their code because they're both built on fsspec, but I figure we can test now, refactor later.

### DIFF
--- a/gridded_etl_tools/utils/store.py
+++ b/gridded_etl_tools/utils/store.py
@@ -622,7 +622,11 @@ class Local(StoreInterface):
         metadata_path = pathlib.Path(self.get_metadata_path(title, stac_type))
         if metadata_path.exists():
             # Generate history file
-            old_mod_time = datetime.datetime.fromtimestamp(metadata_path.stat().st_mtime)
+            old_mod_time = (
+                datetime.datetime.fromtimestamp(metadata_path.stat().st_mtime)
+                .astimezone(datetime.timezone.utc)
+                .replace(tzinfo=None)
+            )
             history_path = (
                 pathlib.Path(self.folder) / "history" / title / f"{title}-{old_mod_time.isoformat(sep='T')}.json"
             )
--- a/gridded_etl_tools/utils/store.py
+++ b/gridded_etl_tools/utils/store.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 from typing import TYPE_CHECKING
 
-if TYPE_CHECKING:
+if TYPE_CHECKING:  # pragma NO COVER
     from .. import dataset_manager
 
 import datetime
@@ -43,7 +43,7 @@ class StoreInterface(ABC):
         self.dm = dm
 
     @abstractmethod
-    def mapper(self, **kwargs: dict) -> collections.abc.MutableMapping:
+    def mapper(self) -> collections.abc.MutableMapping:
         """
         Parameters
         ----------
@@ -55,7 +55,6 @@ class StoreInterface(ABC):
         collections.abc.MutableMapping
             A key/value mapping of files to contents
         """
-        pass
 
     @property
     @abstractmethod
@@ -66,7 +65,6 @@ class StoreInterface(ABC):
         bool
             Return `True` if there is existing data for this dataset on the store.
         """
-        pass
 
     @abstractmethod
     def metadata_exists(self, title: str, stac_type: str) -> bool:
@@ -153,7 +151,7 @@ class StoreInterface(ABC):
             A URL string starting with "s3://" followed by the path to the Zarr.
         """
 
-    def dataset(self, **kwargs: dict) -> xr.Dataset | None:
+    def dataset(self) -> xr.Dataset | None:
         """
         Parameters
         ----------
@@ -167,7 +165,7 @@ class StoreInterface(ABC):
             The dataset opened in xarray or None if there is no dataset currently stored.
         """
         if self.has_existing:
-            return xr.open_zarr(self.mapper(**kwargs))
+            return xr.open_zarr(self.mapper())
         else:
             return None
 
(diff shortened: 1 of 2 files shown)


rbanick (approved):

gridded_etl_tools/utils/store.py
```diff
@@ -153,7 +151,7 @@ def path(self) -> str:
             A URL string starting with "s3://" followed by the path to the Zarr.
         """
 
-    def dataset(self, **kwargs: dict) -> xr.Dataset | None:
+    def dataset(self) -> xr.Dataset | None:
```
ohsqueezy, line 154: The kwargs provides a shortcut to getting a refreshed mapper and therefore non-cached dataset but a quick grep shows it should be safe to remove if desired. If so, the docstring needs updating.

gridded_etl_tools/utils/store.py
```diff
@@ -266,9 +263,12 @@ def path(self) -> str:
             return f"s3://{self.bucket}/datasets/{self.dm.key()}.zarr"
 
     def __str__(self) -> str:
+        # TODO: Is anything relying on this? It's not super intuitive behavior. If this is for debugging in a REPL, it
```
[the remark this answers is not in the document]
ohsqueezy: A quick grep of tools and managers shows it should be safe to change but don't we want `__str__` to be the path? The value is not meant to be evaluated directly it's just for printing the store path in the logs. I was going by the discussion here at the time https://stackoverflow.com/questions/1436703/what-is-the-difference-between-str-and-repr

gridded_etl_tools/utils/store.py
```diff
@@ -418,7 +421,7 @@ class IPLD(StoreInterface):
     used to write new data to IPFS and generate a new recursive hash.
     """
 
-    def mapper(self, set_root: bool = True, refresh: bool = False, **kwargs: dict) -> ipldstore.IPLDStore:
+    def mapper(self, set_root: bool = True, refresh: bool = False) -> ipldstore.IPLDStore:
```
ohsqueezy, line 424: kwargs section needs to be removed from the docstring

gridded_etl_tools/utils/store.py
```diff
@@ -43,19 +43,18 @@ def __init__(self, dm: dataset_manager.DatasetManager):
...
 
     @abstractmethod
-    def mapper(self, **kwargs: dict) -> collections.abc.MutableMapping:
+    def mapper(self, **kwargs) -> collections.abc.MutableMapping:
         """
         Parameters
         ----------
         **kwargs : dict
-            Implementation specific keywords
+            Implementation specific keywords. TODO: standardize interface across implementations
```
ohsqueezy, line 51: I understand inherited function signatures need to be compatible with the parent to satisfy liskov substitution but my understanding is using kwargs as a way to satisfy it is acceptable practice in Python? I'm thinking of something like this for example https://stackoverflow.com/a/54155637

gridded_etl_tools/utils/store.py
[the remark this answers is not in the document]
chrisrossi: It's not obvious to me that stringifying a store should only yield a path. If you want to print the path in the logs, you can access the path directly as `store.path`. Generally speaking, unless there is an obvious two way conversion between a string and another type, like there is, for instance, with numeric types, I wouldn't use `__str__`. If the goal is simply to make something human readable, `__repr__` is the special method for that, and should return something like: "S3('s3://bucket/path/to.zarr')"

gridded_etl_tools/utils/store.py
chrisrossi: That works, but I would discourage it, because the point of having an interface is to provide polymorphism, which means that calling code can/should be written in a way that is unaware of which implementation of an interface it is using. In order to access those extra keyword arguments, the caller has to be aware that it has a more specific implementation than the generic interface. It's best to avoid doing that if you can. 

In this particular instance, I could imagine bringing `refresh` up to interface level, since I can imagine that being appropriate universally. I can also imagine moving `set_root` from `mapper` to the constructor for `IPLD`, since the constructor is not part of `StoreInterface` and doesn't need to be generic, and we can probably make that decision when instantiating the store.

gridded_etl_tools/utils/store.py
chrisrossi: This was an artifact of an earlier draft where I got rid of `**kwargs` from `mapper`, and then realized they were being used. I should add `**kwargs` back here, too. I would eventually like to see them phased out, but in a deliberate, thoughtful, and tested way.

chrisrossi: 62db238

gridded_etl_tools/utils/store.py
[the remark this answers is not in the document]
ohsqueezy: Alright I guess that kind of contradicts the consensus at that link which is saying to use `__str__` to give a human readable representation but I agree that store doesn't need or wouldn't be expected to have a string conversion method

gridded_etl_tools/utils/store.py
ohsqueezy: Yeah that design seems better but I guess there will still be different return types for mapper so for example you can call set_root on an IPLD mapper but not an S3 mapper

gridded_etl_tools/utils/store.py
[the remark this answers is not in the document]
chrisrossi: There isn't always a universal consensus on best practices for software development. This is one of those things where I've shared how I look at it, but you're free to ignore my feedback.

gridded_etl_tools/utils/store.py
chrisrossi: Ideally the `Mapper` interface would get standardized as well, but one step at a time, and we don't always achieve ideal. (We mostly don't.) The more we keep things like this in mind, though, the better designs we'll make when writing new code.

PR #145, opened 2024-02-27
Make profile of S3 store configurable

rbanick:


gridded_etl_tools/utils/store.py
```diff
@@ -224,7 +224,7 @@ def __init__(self, dm: dataset_manager.DatasetManager, bucket: str):
             raise ValueError("Must provide bucket name if parsing to S3")
         self.bucket = bucket
 
-    def fs(self, refresh: bool = False) -> s3fs.S3FileSystem:
+    def fs(self, refresh: bool = False, profile: str = None) -> s3fs.S3FileSystem:
```
ohsqueezy, line 227: type should be `str | None`

ohsqueezy (approved):

gridded_etl_tools/utils/store.py
```diff
@@ -233,14 +233,20 @@ def fs(self, refresh: bool = False) -> s3fs.S3FileSystem:
...
 
         Returns
         -------
         s3fs.S3FileSystem
             A filesystem object for interfacing with S3
         """
         if refresh or not hasattr(self, "_fs"):
-            self._fs = s3fs.S3FileSystem()
+            if profile:
+                self._fs = s3fs.S3FileSystem(profile=profile)
```
chrisrossi, line 247: What would happen if we passed `profile=None` here? Do we need the if/else?

rbanick: You’re right, this would simplify things

rbanick: Zapped!

PR #146, opened 2024-02-27
fixing broken post parse QC test reading from S3-based JSONs

rbanick:


gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -811,11 +811,11 @@ def zarr_json_to_dataset(self, zarr_json_path: str = None, decode_times: bool =
         if not zarr_json_path:
             zarr_json_path = str(self.zarr_json_path())
 
-        dataset = xr.open_dataset(
-            "reference://",
-            engine="zarr",
-            chunks={},
-            backend_kwargs={
+        input_kwargs = {
```
chrisrossi, line 814: Curious, any reason you prefer to construct the dictionary rather than pass directly?

rbanick: Since this method is where the JSON rubber hits the Xarray road we often enough find ourselves throwing debuggers in this method and tweaking parameters. Tweaking the full set of parameters is a PIA especially when debuggers don’t play nicely with indents so having a dict will make it easier to debug going forward.

PR #150, opened 2024-03-06
chore: squash warnings

chrisrossi:
Fix all warnings originating in our code. Update pytest config to treat warnings originating in our code as errors to help us stay on top of deprecated APIs and other such things.

### DIFF
--- a/gridded_etl_tools/__init__.py
+++ b/gridded_etl_tools/__init__.py
@@ -1,6 +1,0 @@
-from pkg_resources import get_distribution, DistributionNotFound
-
-try:
-    __version__ = get_distribution(__name__).version
-except DistributionNotFound:  # pragma NO COVER
-    pass  # package is not installed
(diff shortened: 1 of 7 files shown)


gridded_etl_tools/__init__.py
chrisrossi, line 1: This was deprecated and seems to have been redundant with `setuptools_scm`. I haven't noticed any impact from removing it. I've tested installing with this change and pip still finds the version just fine.

tests/pytest.ini
```diff
@@ -1,2 +1,7 @@
 [pytest]
-log_level=INFO
+log_level=WARN
```
chrisrossi, line 2: Attempting to make the output from the system tests less verbose, as it there can be an unreal amount of scrolling up required to find the actual error. This helps some, but it seems the same logging is winding up on stderr which is getting captured and displayed by pytest as well. But now it's only printing it once instead of twice, which is an improvement.

rbanick (approved):

tests/pytest.ini
rbanick: Appreciate this, agree the scrolling is maddening

PR #157, opened 2024-03-19
refactor extractor to use asyncio

rbanick:


gridded_etl_tools/utils/extractor.py
```diff
@@ -10,75 +10,64 @@
...
-    Returns
-    -------
-    bool
-        `True` if any of the batch calls returns `True`, `False` otherwise
-    """
-    # Success remains Trueuntil the first unsuccessful request is completed
-    success = True
+class Extractor(ABC):
+
+    def __init__(self, dm: dataset_manager.DatasetManager, concurrency_limit: int = 8):
```
rbanick, line 29: no default to number of cores - 1?

gridded_etl_tools/utils/extractor.py
```diff
@@ -10,75 +10,64 @@
...
 import pathlib
 import typing
 import ftplib
 import re
 import time
-import multiprocessing
 import logging
 
 
 log = logging.getLogger("extraction_logs")
```
rbanick, line 24: will this get passed through to Prefect as things stand or do we need to add it to Prefect’s loggers?

gridded_etl_tools/utils/extractor.py
```diff
@@ -10,75 +10,64 @@
...
 
-    # Thread count is one less than the amount of CPU available, unless there is only 1 CPU
-    thread_count = max(1, multiprocessing.cpu_count() - 1)
+        Returns:
+            bool: True if all of the jobs succeeded, False otherwise.
+        """
+        nest_asyncio.apply()
+        coros = [asyncio.to_thread(self.request, *job_arg) for job_arg in batch]
+        results = asyncio.run(self.async_gather_with_semaphore(coros))
+        return all(results)
```
rbanick, line 46: I feel like we need some messaging attached to success/failure.

gridded_etl_tools/utils/extractor.py
```diff
@@ -311,11 +298,11 @@ def request(
         log.info(f"Downloading remote file {source} to {output}")
         with open(output, "wb") as fp:
             try:
-                # Use a semaphore to limit the number of simultaneous downloads to 1 even in multithreaded
-                # environments. This is either a requirement of ftplib or a common requirement of FTP servers.
-                with self.semaphore:
-                    log.info("Using a single thread as a requirement of FTP even if multiple threads are available")
-                    self.ftp.retrbinary(f"RETR {source}", fp.write)
+                # need separate FTP for each download to take advantage of concurrency
```
rbanick, line 301: so will these actually run concurrently?

gridded_etl_tools/utils/extractor.py
```diff
@@ -10,75 +10,64 @@
...
+        Returns:
+            list[bool]: A list of the results of the coroutines
+        """
 
-        log.info(f"Downloading took {(time.time() - start_downloading) / 60:.2f} minutes")
-    else:
-        log.info("No requests prepared to be submitted, please check batch_processor is functioning properly.")
-        success = False
+        async def sem_coro(coro):
+            async with self.semaphore:
```
rbanick, line 60: how does the Sempahore work in this instance? It detects if the specified concurrency_limit is reached and holds?

gridded_etl_tools/utils/extractor.py
```diff
@@ -10,75 +10,64 @@
...
+                return await coro
 
-    # If every request returned False, log a message
-    if not success:
-        log.info("At least one request for data was unsuccessful.")
+        return await asyncio.gather(*(sem_coro(c) for c in coros))
 
-    return success
+    @abstractmethod
+    def request(self, *args, **kwargs) -> bool:
```
rbanick, line 66: Should include a short docstring detailing expectations for child implementations.

gridded_etl_tools/utils/extractor.py
eschechter: Nope, the number of cores is completely irrelevant for this implementation. The main limitation is whether your target system gets angry at you. In the case of FTPs, that seems to happen for anything above 2. For other extractors, much higher might be possible

gridded_etl_tools/utils/extractor.py
eschechter: They will if you set the `concurrency_limit` to be more than 1. It's unclear if that's advisable for FTPs though, since they seem to be finicky with any sort of parallelism.

gridded_etl_tools/utils/extractor.py
eschechter: Yup, if at the `concurrency_limit`, it blocks until another coroutine releases the semaphore.

gridded_etl_tools/utils/extractor.py
eschechter: In other words, the code fundamentally works, but might be slower or less reliable than just running the code linearly.

ohsqueezy: I think I tried multiple connections and it was either slower or didn't work. I was mostly looking at [this thread](https://stackoverflow.com/questions/2549829/using-ftplib-for-multithread-uploads) at the time. If concurrency is 1 by default, then maybe it's fine to leave it as an option though.

gridded_etl_tools/utils/extractor.py
ohsqueezy: For initial parses on Copernicus API, we want to immediately launch all thousands of requests so that they get registered in the queue. In that case, I'm guessing we'll need to override the default of 8 in ERA5's extract code.

ohsqueezy (approved):
I'll approve but keep in mind anything that requests more than 8 files from Copernicus will now be slower if I'm understanding correctly

PR #161, opened 2024-04-12
fix: array regression (option 2)

### LINKED ISSUE #162: Latest version of Xarray doesn't like Zarr chunks spanning multiple Dask chunks.
With the latest version of xarray, the test `test_append_parse` for GFS in the managers repo fails with:

```
FAILED tests/managers/gfs/test_gfs.py::test_append_parse - ValueError: Specified zarr chunks encoding['chunks']=(10, 104, 103, 16) 
for variable named 'prate' would overlap multiple dask chunks ((1, 1), (6,), (103, 103, 103, 103, 103, 103, 103), (16, 16, 16, 16...
```

Although the test failure is in managers, the fix should happen here, in tools, and include a regression test that exposes the problem.

chrisrossi:
Use requested dask chunks for appends.

Fixes #162


chrisrossi:
@rbanick The new test you added is something we need, as there should absolutely be a unit test for that, but it's not the regression test referred to in #162 as it doesn't actually trigger the error we're trying to fix. Because that error is being raised by xarray, we likely can only get a true regression test from a system or integration test. I was going to look into writing or modifying a system test myself, but if you want to take a stab at it, you are welcome.

eschechter (approved):

PR #165, opened 2024-05-07
HTTP Extractor

rbanick:


gridded_etl_tools/utils/extractor.py
```diff
@@ -72,6 +72,38 @@ def request(self, *args, **kwargs) -> bool:
...
+        if not hasattr(dm, "session"):
+            dm.get_session()
+
+    def request(self, remote_file_path: str, local_file_path: str) -> bool:
+        """
+        Request a file from an HTTP Server and save it to disk
+        Requires an active session within the dataset manager
+        """
+        self.dm.info(f"Downloading {local_file_path}")
+        fil_in_mem = self.dm.session.get(remote_file_path)
```
chrisrossi, line 112: Are we sure the files will fit in memory? This could stream so we don't have to load the while into memory at once. If we're sure the files will fit in RAM easily, then maybe we don't need to do that.

chrisrossi (approved):
Looks ok to me.

ohsqueezy (approved):

gridded_etl_tools/utils/extractor.py
rbanick: Good idea, I've tweaked accordingly

PR #172, opened 2024-05-20
Convert unusual time values w/in post parse QC

rbanick:


gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -1603,6 +1603,35 @@ def raw_file_to_dataset(self, file_path: str) -> xr.Dataset:
...
+        time_values = np.atleast_1d(ds[time_dim].values)
+
+        # Convert any raw time values in esoteric formats that break the binary search comparison logic.
+        # NOTE the expectation is that the `convert_raw_times` function is defined w/in the ETL manager
+        if type(time_values[0]) is not np.datetime64:
+            time_values = self.convert_raw_times_to_numpy_times(time_values)
+
+        return time_values
+
+    def convert_raw_times_to_numpy_times(self, raw_times: np.array) -> np.array:
```
eschechter, line 1631: in the type hints, `np.ndarray`, not `np.array`

gridded_etl_tools/utils/zarr_methods.py
```diff
@@ -1603,6 +1603,35 @@ def raw_file_to_dataset(self, file_path: str) -> xr.Dataset:
         else:
             raise ValueError('Expected either "file" or "s3" protocol')
 
+    def convert_raw_times_to_comparable_times(self, ds: xr.Dataset, time_dim: str) -> np.ndarray | list:
```
eschechter, line 1606: Ideally, the type hint would specify one of the two current possible return types, since the two have very different methods that can be called on them. Is there any way this function can return a `list`? My hunch is it can only return `np.ndarray`

rbanick: The downstream implementation returns a list but I've changed that in the responsible manager so now we can verifiably return a `np.ndarray`. Better that way.

eschechter (approved):

PR #174, opened 2024-05-21
Separate out transformations of raw data on disk and in-memory Xarray Datasets

rbanick:


gridded_etl_tools/utils/convenience.py
```diff
@@ -348,6 +348,26 @@ def date_range_to_string(self, date_range: tuple) -> tuple[str, str]:
             datetime.datetime.strftime(date_range[1], "%Y%m%d%H"),
         )
 
+    def strings_to_date_range(self, date_range: tuple, parse_string: str = "%Y%m%d%H") -> tuple[str, str]:
```
ohsqueezy, line 351: The return type should be `tuple[datetime.datetime, datetime.datetime]`

gridded_etl_tools/utils/convenience.py
```diff
@@ -348,6 +348,26 @@ def date_range_to_string(self, date_range: tuple) -> tuple[str, str]:
...
+        """
+        Convert a tuple of parseable strings to datetime objects. Necessary for Xarray metadata parsing.
+
+        Parameters
```
ohsqueezy, line 355: The input parameters and return value should be switched

gridded_etl_tools/utils/publish.py
```diff
@@ -1,551 +1,55 @@
...
-        kwargs : dict
-            Keyword arguments specified by the user
+        publish_dataset : xr.Dataset
+            A dataset containing all records to publish, either as an initial dataset or an update to an existing one
 
         Returns
         -------
         bool
-            Flag indicating if new data was / was not parsed
+            Flag indicating if new data was / was not successfully parsed
```
ohsqueezy, line 52: I guess we actually don't return a value anymore

gridded_etl_tools/utils/transform.py
```diff
@@ -0,0 +1,727 @@
...
+        outfile_path
+            A custom string path for the final, merged Zarr JSON.
+            Defaults to None, in which case Zarr JSONs are output to self.zarr_json_path.
+        """
+        self.zarr_json_path().parent.mkdir(mode=0o755, exist_ok=True)
+        # Generate a multizarr if it doesn't exist. If one exists, overwrite it unless directed otherwise.
+        if not self.zarr_json_path().exists() or force_overwrite:
+            start_kerchunking = time.time()
+            # Prepapre a list of zarr_jsons and feed that to MultiZarrtoZarr
+            if not hasattr(self, "zarr_jsons"):
```
ohsqueezy, line 73: If zarr_jsons is read and written to from outside of this class, we probably should give it a default value at construction time

ohsqueezy (approved):
I haven't looked at managers yet but tools looks OK to merge. 

Distinguishing between publish and transform seems good since it makes sense that a client could want to publish a single transformation multiple ways, or maybe perform multiple transformations using one publish.

I wonder if we should further distinguish between remote and local transformation classes. Remote transformations have made the distinction between transformation and extraction unclear. Remote transformations seem like they need to be arguments to extract, but local transformations can happen independent from extract.

gridded_etl_tools/dataset_manager.py
```diff
@@ -283,10 +283,15 @@ def extract(self, date_range: tuple[datetime.datetime, datetime.datetime] | None
             )
         self.new_files = []
 
-    def transform(self):
+    def transform_data_on_disk(self):
```
ohsqueezy, line 286: After looking at managers I'm wondering if this function could be named "virtualize"

gridded_etl_tools/dataset_manager.py
```diff
@@ -297,6 +302,38 @@ def transform(self):
         # from Dask
         self.create_zarr_json()
 
+    def transform_dataset_in_memory(self) -> xr.Dataset:
```
ohsqueezy, line 305: I'm actually concerned about this function returning a value now that I look at managers. Previously we just transformed the ETL object internally/in-place. 

I think returning values is one way to make ETL objects more versatile, but the advantage of not returning the dataset here was a client using a pre-made object like CPCPrecipUS did not need to know about xarray.

I think in general we want to preserve a concept of a manager which is able to run an ETL just by running functions. And especially in this case it seems counter-intuitive to give the client an unfinished dataset and expect the client to do the next step with it.

gridded_etl_tools/utils/transform.py
rbanick: For the time being we've done so for S3 Extractors in the extractor init. @eschechter and I have gone back andforth on this quite a bit though as it's messy. We probably need to figure out a better solution than `zarr_jsons` in the long run

https://github.com/Arbol-Project/gridded-etl-tools/blob/main/gridded_etl_tools/utils/extractor.py#L136

gridded_etl_tools/dataset_manager.py
rbanick: @eschechter could you chime in here, since we very specifically spoke about this?

gridded_etl_tools/utils/transform.py
ohsqueezy: This seems related to another comment I wanted to discuss

> I wonder if we should further distinguish between remote and local transformation classes. Remote transformations have made the distinction between transformation and extraction unclear. Remote transformations seem like they need to be arguments to extract, but local transformations can happen independent from extract.

If we for example pass the transform object to the extract object then it seems like we can let transform be in charge of zarr_jsons

rbanick: How do you propose to pass Transform objects to Extract objects? Can you give a pseudocode example?

In general I am strongly skeptical of any attempt to re-combine Transform with either Extract or Publish -- we've spent a lot of effort achieving functional separation as it makes the code a lot easier to write, maintain, and follow. So there'd have to be a good reason born out in actual practice.

gridded_etl_tools/dataset_manager.py
ohsqueezy: It seems OK to me if the Zarr JSON is referred to as the virtual Zarr even when it's not loaded into memory, and isn't that how `zarr_json_to_dataset` describes the virtual Zarr?
```
def zarr_json_to_dataset(self, zarr_json_path: str = None, decode_times: bool = True) -> xr.Dataset:
     """
     Open the virtual zarr at `self.zarr_json_path()` and return as a xr.Dataset object after applying
     any desired postprocessing steps
```
But I should have mentioned the reason I bring it up is that having both `transform_data_and_disk` and `transform_dataset_in_memory` start with the word transform is confusing at least to me. One rewrites files on disk and creates JSON, and the other returns an xr.Dataset. I think when we refactor transform to be its own class we'll further improve this situation, but clearer distinction between the two new transform function names might be something good to have for now.

gridded_etl_tools/utils/transform.py
ohsqueezy: For sure, what I'm getting at is the same idea. We want transform and extract to be separate, but we see with `zarr_jsons` and I think other properties as well that they are currently tied to each other.

So maybe something like the following could work because it would leave `zarr_jsons` internal to the Transform class.

```
my_etl = CPCPrecipUS()
transformation = Transform(my_etl, Transformation.REMOTE_KERCHUNK)
extract = S3Extractor(my_etl, transformation) # transformation.zarr_jsons available to S3Extractor
```

rbanick: The critical thing here is that `kerchunk` is used for Extract (remotely) and transformation (locally). This means the function and its dependencies (i.e. `zarr_jsons`) pop up in both places and have to be accommodated as such.

So in this case Remote Kerchunk isn't really a tansformation, it's an extraction that makes `transform_data_on_disk` moot since the extracted files are the final files. That seems more logical and easy to follow then instantiating a Remote "transformation" that is not.

rbanick: Looking again at this I'm struggling to see what the marginal gains of this approach are relative to re-writing all our ETLs and transformation stack, which is what it would take. Why is passing transform to extract so advantageous, outside of managing the `zarr_jsons` instance variable more elegantly?

ohsqueezy: To clarify, I'm not proposing we do this in this PR. For this PR, I just think we should consider creating zarr_jsons in the Transform class instead of Extractor, dropping `hasattr`, and then just continuing to let Extract modify zarr_jsons freely for now.

For the approach in general, my reasoning is we're working on refactoring away from DM being a mix-in, so there will need to be a new way of communication between classes. We're taking a step with basic composition, so extract will be able to access zarr_jsons using `self.dm.transform.zarr_jsons`. That's probably an improvement, but I think the goal for any class is for it to work fully independently so that it can be easily reused. If we can further move away from needing `self.dm`, we can work toward having a Transform class that can be applied generally to any data.

About the remote kerchunk, I'm confused because S3Extractor currently calls a request a transformation.

```
def request:
     """
     Transform a remote S3 climate file into a JSON file and add it to the given `DatasetManager` object's internal list of Zarr JSONs.
```

rbanick: Let me rename that docstring, it's confusing, you're right

PR #177, opened 2024-06-06
Expanded post parse QC

rbanick:


gridded_etl_tools/utils/publish.py
```diff
@@ -793,35 +793,147 @@ def check_written_value(
...
         Returns
         ----------
         orig_ds
-            The original dataset, unformatted
+            The original dataset, formatted minimally to allow comparison with the production dataset
         """
-        # Randomly select an original dataset
-        raw_ds, orig_file_path = self.binary_search_for_file(coords[self.time_dim], time_dim=self.time_dim)
+        # Apply any custom filters to the input files. Returns all input files if no custom filter defined.
+        possible_files = self.custom_file_filter(list(self.input_files()))
```
ohsqueezy, line 811: Does this change between each coordinate check? If not, we probably want this and the other filters in this function to happen before iterating through the coordinates.

gridded_etl_tools/utils/publish.py
```diff
@@ -793,35 +793,147 @@ def check_written_value(
...
+                coords["forecast_reference_offset"],
+                time_dim="forecast_reference_offset",
+                possible_files=ensemble_filtered_original_files,
             )
 
-        # Reformat the dataset such that it can be selected from equivalently to the prod dataset
-        orig_ds = self.reformat_orig_ds(raw_ds, orig_file_path)
         return orig_ds
 
+    def custom_file_filter(self, original_files: tuple[str]) -> tuple[str]:
```
ohsqueezy, line 848: I guess this is somewhat subjective, but it seems redundant to call this a custom filter. If it's the object's filter function, it's not really custom, even if it's a modification of the parent's function. My expectation would be a custom filter would be a function that is passed in as an argument.

gridded_etl_tools/utils/publish.py
```diff
@@ -793,35 +793,147 @@ def check_written_value(
...
+        ensemble_filtered_original_files : tuple[str]
+            A list of raw files, filtered to only contain files that contain the specified ensemble
+        """
+        ensemble_string = "ensemble-" + str(raw_ds["ensemble"].values[0]) + "_"
+        ensemble_filtered_original_files = [fil for fil in original_files if ensemble_string in str(fil)]
+        return ensemble_filtered_original_files
+
     def binary_search_for_file(
-        self, target_datetime: np.datetime64, time_dim: str, possible_files: list[str] | None = None
+        self, target_datetime: np.datetime64, time_dim: str, possible_files: list[str]
```
ohsqueezy, line 936: I realize I'm late to the party with this comment, but I'm not sure a binary search is the best solution here. Wouldn't it be better to just choose the files beforehand at the same time the coordinates are chosen? Then we could just iterate over the list of coordinates+files. If the pure randomness is a problem, we could use a different version of randomization where we go in order over the sorted list of files at the start of the post parse check, randomly choosing ones to check, and randomly choosing a coordinate or coordinates in the files.

gridded_etl_tools/utils/publish.py
```diff
@@ -702,6 +710,92 @@ def post_parse_quality_check(self, checks: int = 100, threshold: float = 10e-5):
...
+        ----------
+        production_date_range : tuple[datetime.datetime, datetime.datetime]
+            The date range for the production dataset
+        orig_file_path : pathlib.Path
+            The path to the original file, on disk or remotely
+
+        Returns
+        -------
+        bool
+            _description_
```
ohsqueezy, line 767: Is _description_ a reference to something

gridded_etl_tools/utils/publish.py
```diff
@@ -702,6 +710,92 @@ def post_parse_quality_check(self, checks: int = 100, threshold: float = 10e-5):
...
+        if (
+            production_date_range[0]
+            <= self.numpydate_to_py(orig_ds[self.time_dim].values[0])
+            <= production_date_range[1]
+        ):
+            return True
+        else:
+            return False
+
+    def custom_file_filter(self, original_files: tuple[str]) -> tuple[str]:
```
ohsqueezy, line 780: I still think we don't need custom_file_filter. The convention we're already using where we override the existing function is better design-wise.

```py
class Publish:

    def filter_search_space(self):
        return ["file0.nc4", "file1.nc4", "file2.nc4"]

class MyETL:

    def filter_search_space(self):
        possible_files = super().filter_search_space()
        return possible_files[1:] // the customization is to always remove the first file
```

ohsqueezy (approved):

gridded_etl_tools/utils/publish.py
rbanick: OK this makes sense, I'm implementing it so

PR #178, opened 2024-06-10
introduce convenience function to get and return next possible date as date range

rbanick:


gridded_etl_tools/utils/convenience.py
```diff
@@ -382,6 +382,35 @@ def get_newest_file_date_range(self) -> datetime.datetime:
         """
         return self.get_date_range_from_file(list(self.input_files())[-1])
 
+    def get_next_date_as_date_range(self) -> tuple[datetime.datetime, datetime.datetime]:
```
ohsqueezy, line 385: If we're adding this, do we want to also add a get_next_date function which this function would call. Without it, if we want to get the next date programmatically using this function we would have to do get_next_date_as_date_range()[0]

ohsqueezy (approved):

gridded_etl_tools/utils/convenience.py
rbanick: Nice, me gusta. I made this into a property

PR #192, opened 2024-08-20
Extractor: create get_hrefs method

rbanick:


gridded_etl_tools/utils/extractor.py
```diff
@@ -116,6 +117,30 @@ def request(self, remote_file_path: str, local_file_path: str) -> bool:
...
+        -------
+        list[str]
+            A list of links, in string format, from the specified URL
+        """
+        self.dm.info("Getting links from: " + url)
+        page = self.dm.session.get(url, timeout=10)
+        html = page.content.decode("utf-8")
+        tree = etree.parse(StringIO(html), parser=etree.HTMLParser())
+        refs = tree.xpath("//a")
+        return [
```
ohsqueezy, line 138: I think separating out the call to set would help with readability here

gridded_etl_tools/utils/extractor.py
```diff
@@ -116,6 +117,27 @@ def request(self, remote_file_path: str, local_file_path: str) -> bool:
...
+        list[str]
+            A list of links, in string format, from the specified URL
+        """
+        self.dm.info("Getting links from: " + url)
+        page = self.dm.session.get(url, timeout=10)
+        html = page.content.decode("utf-8")
+        tree = etree.parse(StringIO(html), parser=etree.HTMLParser())
+        refs = tree.xpath("//a")
+        href_links = set([link.get("href", "") for link in refs])
+        return [link for link in href_links if "https" not in link and "mailto" not in link]
```
ohsqueezy, line 139: Between CIDs and obscure scientific naming schemes it's actually possible these strings would appear in a link name... I think `not link.startswith("https://")` and `not link.startswith("mailto:")` would be OK but also I'm about to leave another comment about using BeautifulSoup.

Also excluding https from the results is confusing as default behavior for a library function

ohsqueezy:
I think it should be up to the user to specify the links that are excluded from the results. We may end up using external HTTPS links in another dataset, for example. That could be done by adding an argument with a list of strings to filter, but with BeautifulSoup there's also the possibility of passing a more generic filter.

Another thing is most people recommend using BeautifulSoup over the built in XML/HTML parser. It parses urls easily and then it allows for stuff like `soup.find_all('a', href=[SOME FILTER])`. So our library could then accept the filter argument. There are a few examples in [the docs](https://beautiful-soup-4.readthedocs.io/en/latest/#the-keyword-arguments)

gridded_etl_tools/utils/extractor.py
```diff
@@ -418,3 +397,35 @@ def batch_requests(self, pattern: str = ".*") -> list[pathlib.PurePosixPath]:
...
+    """
+    Get a list of links from a given URL.
+
+    Parameters
+    ----------
+    session : requests.Session
+        An existing session object to make requests.
+    url : str
+        A URL to scrape links from.
+    filters : tuple[str]
```
ohsqueezy, line 412: I still think filtering against just the beginning of the url string is a bit too specific for a library function, especially considering that BeautifulSoup provides a general filter that could alternatively be passed.

We do in many cases use other filters like "get all *.nc4" or "get everything except *.txt" etc.

In the case of excluding https/mailto, the filter would be a custom function (which can be passed as a lambda since it's a short function)

```
soup.find_all("a", lambda href: not href.startswith("https://") and not href.startswith("mailto://"))
```

So what I'm proposing is that the filter argument to `extractor.get_hrefs` would accept a function rather that a list of strings, so that the following would be the call to get all links that don't start with "https://" or "mailto://".

```
extractor.get_hrefs(session, url, lambda href: not href.startswith("https://") and not href.startswith("mailto://"))
```

ohsqueezy:
I think the get_hrefs function is a good example to use to talk about what this library should look like once we refactor, so I figured I would share some thoughts.

I think the `session` variable is one of the main reasons we would want to consider keeping get_hrefs within the HTTPExtractor class. If HTTPExtractor is a class, we can easily let it handle storing the session, and then we can call `HTTPExtractor.get_hrefs` without having to worry about tracking the session. A new session would be created every time a new `HTTPExtractor` object is created and would last until the object is destroyed.

This would be similar to how `FTPExtractor` handles its connection. A new connection is opened when the FTPExtractor context is opened, and the connection is closed when the context is closed. If we used a module (like `extract.ftp` or something) instead of a class like `FTPExtractor`, we would then force the manager code to handle opening the FTP connection, which would then need to be passed into every other FTP-related function like `cwd` or `find`, etc.

Something related I wanted to mention: I'm just now noticing we nuked a lot of useful things when we moved and undid the move of the pool function. Take a look at what it looked like in https://github.com/Arbol-Project/gridded-etl-tools/commit/e348804b100cb556162495d3d021f78cef47e8ea vs what it looks like now. The docstrings alone lost a lot of good content, and the type hinting is gone.  More importantly though and related to the above is the request function is no longer an argument. It's back to being an abstract function that DM is forced to implement. Regardless of how we refactor, that's one part we definitely want to change.

To summarize, I think removing mix-ins and removing most of the inheritance from DM doesn't necessarily mean that we have to move entirely to using modules over classes.

ohsqueezy:
Maybe we want the the FTPExtractor behavior: destination path is either explicitly given or the working directory is used. It helps further decouple from DM and doesn't require manager changes.

Related side note, I'm seeing now I should either update the behavior or the docs because it's confusing that it says HTTPExtractor uses destination_path but then it's actually local_input_path / destination_path

gridded_etl_tools/utils/extractor.py
```diff
@@ -59,21 +87,31 @@ def pool(self, batch: typing.Sequence[typing.Dict]) -> bool:
...
+        arg
+            A single argument or list/dict of arguments to be passed to the request method
 
         Returns
         -------
         bool
             True if the request was successful, False otherwise.
         """
-        return self.request(**dict_arg)
+        print(arg)
```
eschechter, line 108: print statement should go

eschechter (approved):

PR #193, opened 2024-08-30
Update filtering and scraping in HTTP extractor

ohsqueezy:
The new filter_func argument to the link scraping function accepts a generic function that receives the link href string and returns a boolean indicating whether or not to filter the link out.

The code for link scraping is moved into HTTPExtractor and renamed from get_hrefs to get_links because it now returns the absolute URL of each link instead of literally its href value.

HTTPExtractor creates a requests.session object, freeing up the caller from having to manage one of its own. HTTPExtractor is now required to be created within a context manager, which it uses to automatically open and close the session.

In an effort to make HTTPExtractor.request work smoothly with Extractor.pool when requesting a list of urls to download, Extractor.pool now accepts 1D arguments in addition to 2D ones. 1D arguments are passed one by one to Extractor.pool.

Corresponding tests for HTTPExtractor and Extractor.pool are added to test_extractor. A new testing dependency is added: the "responses" Python package. It mocks responses from HTTP servers, making it possible to test requests for link scraping, filtering, and downloading.

### DIFF
--- a/gridded_etl_tools/utils/convenience.py
+++ b/gridded_etl_tools/utils/convenience.py
@@ -27,12 +27,12 @@ class Convenience(Attributes):
     # BASE DIRECTORIES
 
     def root_directory(self, refresh: bool = False):
-        if refresh or not hasattr(
-            self, "_root_directory"
-        ):  # ensure this is only calculated one time, at the beginning of the script
-            self._root_directory = (
-                pathlib.Path.cwd()
-            )  # Paths are relative to the working directory of the ETL manager, *not* the scripts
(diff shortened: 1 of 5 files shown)


ohsqueezy:
Turned into a bigger change than expected... Most notable changes besides the link scraping and filtering are

- Extractor.pool accepts more types of arguments that can be passed to Extractor.request
- HTTPExtractor acts as a context manager to go along with it creating a session object
- A lot of HTTP request mocking and testing

rbanick (approved):

PR #199, opened 2024-12-09
create apply_compression method to consistently apply compression and…

rbanick:
Inconsistent application of Blosc lz4 compression causes downstream issues. Apply it everywhere in a dedicated method to avoid this issue.


gridded_etl_tools/utils/metadata.py
```diff
@@ -767,15 +793,21 @@ def remove_unwanted_fields(self, dataset: xr.Dataset) -> xr.Dataset:
...
-
         for coord in ["latitude", "longitude"]:
             dataset[coord].attrs.pop("chunks", None)
             dataset[coord].attrs.pop("preferred_chunks", None)
             dataset[coord].encoding.pop("_FillValue", None)
             dataset[coord].encoding.pop("missing_value", None)
-            dataset[coord].encoding["compressor"] = compressor
         dataset[self.data_var].encoding.pop("filters", None)
+
+    def apply_compression(self, dataset: xr.Dataset):
```
ohsqueezy, line 803: I think we want to call this something other than "apply". If I'm understanding, even if this does end up applying compression, it doesn't apply it immediately. And if there already is compression, setting this doesn't do anything, right?

Maybe "set compression encoding" or something like that would make sense. I think this docstring would be a good place to put details about what actually happens, like it will be applied when writing the Zarr if there is no compression already. And if you need to modify compression, use the modify function.

I'm wondering whether it would make sense to raise an error if this is used on a dataset with compression already set. Wouldn't you always need to use modify in that case?

gridded_etl_tools/utils/metadata.py
```diff
@@ -793,6 +825,85 @@ def suppress_invalid_attributes(self, dataset: xr.Dataset):
             elif dataset.attrs[attr] is None:
                 dataset.attrs[attr] = ""
 
+    def modify_array_encoding(
```
ohsqueezy, line 828: This function essentially does three different things: update, insert, and remove. Generally, I'm thinking it would best be three different functions, maybe with each function calling a common function (Python makes this a little awkward, since ideally the common function would also be a private function). I think in this case, since the tests are already well established, it's fine to leave as is.

ohsqueezy (commented):
In the past, we've generally implemented stuff that modifies data on disk as standalone scripts. Potentially, we can do that here without needing to change much. The body of the modify function would be the script, and the tests would be their own test suite that went along with the script.

I'm not necessarily opposed to keeping the function, though.

tests/unit/utils/test_metadata.py
```diff
@@ -1432,3 +1468,239 @@ def test_merge_in_outside_metadata_ipld_stac_timeout(manager_class, fake_origina
...
+    @staticmethod
+    @pytest.fixture
+    def encoding_test_output():
+        """Fixture to create and cleanup output directory for encoding tests"""
+        output_path = pathlib.Path("./tests/unit/utils/output")
+        pathlib.Path.mkdir(output_path, parents=True, exist_ok=True)
+        yield output_path
+        clean_up_input_paths(output_path)
+
+    # @staticmethod
```
ohsqueezy, line 1481: I'm guessing we can remove the commented out parts

rbanick:
I don't think there's necessarily a tension between having a function and a script. My assumption was that actual applications of these functions would occur in standalone scripts, not within ETLs.

Perhaps the functions could exist outside the tools library still; I wanted to write the functions within the main stack so they could be rigorously tested and fully leverage useful member variables of dataset managers. 

Perhaps in the longer term maintenance-type functions should live in their own `utils` file?

Regarding the different applications of the function, fully agreed that it does three different things. I decided against three different functions as it would result in 80% duplicated code within each. I'm open to any ways you can see to square this circle.

ohsqueezy:
For the modify encoding function, something that could work pretty smoothly is if we leave the function as is and make it "private" by adding an underscore at the front of the name like `_modify_array_encoding`. This convention comes with some official [support in Python](https://docs.python.org/3/tutorial/classes.html#private-variables).

Then we can add simple wrapper functions like insert, update, and remove that call the common function.

For the maintenance functions, a separate module could be a good way to keep functions like this and maybe others like rollback in the same repo. Our rule could be something like the maintenance module can call utils but not the other way around.

gridded_etl_tools/utils/metadata.py
rbanick: I've broken into two functions: update and remove. Insert and update use the exact same code so it feels excessive to treat them as if they're separate.

ohsqueezy (approved):

PR #205, opened 2025-03-13
Zarr v3, remove IPLD

rbanick:


ohsqueezy:
Update - the main things happening in this PR are:

* Upgrade to the Zarr 3.0 library. The Zarr, Xarray, Kerchunk, and Dask libraries are the major package bumps, but a lot of other packages are bumped as well.

* Maintain Zarr2 backward compatibility by outputting Zarr2 format Zarrs, with an option to write to Zarr3. This requires using a separate compression object for either Zarr2 or Zarr3 output (both objects implement the same compression algorithm). It also requires separate metadata write functions because the metadata file formats differ between Zarr2 and Zarr3.

* Clean out unused IPFS-related code.

* Force use of naive datetimes to maintain compatibility between all datetime objects in the codebase.

* Set `decode_timedelta` parameter when opening Zarrs, which will soon be a requirement of xarray.

gridded_etl_tools/utils/publish.py
```diff
@@ -150,34 +146,44 @@ def to_zarr(self, dataset: xr.Dataset, *args, **kwargs):
...
-                # a Zarr is opened during writing, there will be indicators that show the data is being edited.
-                self.info("Writing metadata before writing data to indicate write is in progress.")
-                if self.store.has_existing:
-                    update_attrs = {
-                        "update_in_progress": True,
-                        "update_is_append_only": dataset.get("update_is_append_only"),
-                        "initial_parse": False,
-                    }
+            # Determine Zarr format
+            if "zarr_format" in kwargs:
```
eschechter, line 150: I think we `ValueError`  in the case that zarr_format is in the kwargs. That's supposed to be determined by the manager, and I can't think of a case where we'd want that to be overwritten

gridded_etl_tools/dataset_manager.py
```diff
@@ -95,9 +92,10 @@ def __init__(
         skip_pre_parse_nan_check: bool = False,
         skip_post_parse_qc: bool = False,
         skip_post_parse_api_check: bool = False,
-        encryption_key: str = None,
+        encryption_key: str | None = None,
         use_compression: bool = True,
         dry_run: bool = False,
+        output_zarr3: bool = False,
```
eschechter, line 98: I think we should  very aggressively be checking if the user is setting `output_zarr3` on an existing v2 zarr, or vice versa. I don't fully trust `zarr` to raise an appropriate exception and not brick our dataset, as it's had issues with that in the past

gridded_etl_tools/utils/metadata.py
```diff
@@ -670,12 +633,24 @@ def encode_vars(self, dataset: xr.Dataset) -> xr.Dataset:
                 "dtype": self.data_var_dtype,
                 "units": self.unit_of_measurement,
                 "_FillValue": self.missing_value,
-                # deprecated by NUG but maintained for backwards compatibility
                 "missing_value": self.missing_value,
-                "chunks": tuple(val for val in self.requested_zarr_chunks.values()),
-                "preferred_chunks": self.requested_zarr_chunks,
             }
         )
+        # More recent versions of Xarray + Dask choke when updating with pre-chunked update datasets,
```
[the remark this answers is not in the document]
ohsqueezy: I added a note in `Publish.parse`. Just to make sure - this code was intended to be checked into main right? It was already on the branch when I took over, and I haven't done any special testing for it.

gridded_etl_tools/utils/metadata.py
```diff
@@ -371,16 +353,13 @@ def create_stac_item(self, dataset: xr.Dataset):
         properties_dict["start_datetime"] = self.numpydate_to_py(dataset[self.time_dim].values[0]).isoformat() + "Z"
         properties_dict["end_datetime"] = self.numpydate_to_py(dataset[self.time_dim].values[-1]).isoformat() + "Z"
         properties_dict["updated"] = (
-            datetime.datetime.utcnow().replace(tzinfo=datetime.timezone.utc).isoformat()[:-13] + "Z"
+            datetime.datetime.now(tz=datetime.timezone.utc).replace(tzinfo=None).isoformat() + "Z"
```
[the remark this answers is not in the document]
ohsqueezy: I changed it, but the format is tested and passes whether timespec is set or not for some reason. Do we set this globally somewhere? I verified that this line is running correctly because it does fail when set to `timespec="minutes"`.

ohsqueezy:
Ready for re-review. This now checks that any existing Zarr format is the same as the output format before running a parse.

docs/metadata_standard.md
```diff
@@ -68,28 +67,16 @@ A full accounting of STAC's tags and standards is beyond the scope of this READM
...
+Integrating STAC with Zarrs required some additional tweaks to our architecture. The most notable is referencing `zmetadata` under each Item's `assets` instead of the raw `data`. This reflects that the Zarr "files" we open are actually consolidated JOSN metadata referencing all the individual chunks and [aligns with community best practices](https://github.com/radiantearth/stac-browser/issues/44). For the same reason we assign the `zmetadata` the role of `zarr-consolidated-metadata`.
 
 The Pangeo community has been further developing a `datacube` extension for STAC that better describes N-dimensional datasets (see [discussion here](https://discourse.pangeo.io/t/stac-and-earth-systems-datasets/1472)). This extension is powerful but requires additional work to implement, maintain, and read. Given that all of our N-dimensional datasets in practice only have a time dimension we elected not to implement this extension, for now at least.
```
rbanick, line 78: Not related to this PR but let's drop this paragraph it's old and confusing

gridded_etl_tools/utils/metadata.py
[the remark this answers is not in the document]
rbanick: I believe it was something I uncovered during my initial Zarr3 push but don't have a clear memory of it as it was one of many things. To be sure could you create unit tests for it?

gridded_etl_tools/utils/publish.py
```diff
@@ -60,18 +63,25 @@ def parse(self, publish_dataset: xr.Dataset):
...
-                    # data. If rebuild is requested and there is existing data, but allow overwrite is not set, do not
-                    # start parsing and issue a warning.
+                    # has been set, write a new Zarr, overwriting any existing data.
+                    # If rebuild is requested and there is existing data, but allow overwrite is not set, do not start
+                    # parsing and issue a warning.
                     if self.store.has_existing and not self.rebuild_requested:
+
+                        # If zarr.json is present, the format is considered 3. Otherwise, it is considered format 2.
+                        # os.path.join works for both S3 and local.
+                        if self.store.fs().exists(os.path.join(self.store.path, "zarr.json")):
```
rbanick, line 79: Since `output_zarr3` is an init parameter I think this check should take place within `init` itself. I understand the logic of keeping it close to the actual act of publishing data but I think we should colocate parameters and their checks.

PR #207, opened 2025-04-02
Synchronize Zarr v2 metadata present in a Zarr v3 with the latest Zarr v3 Metadata

rbanick:


tests/unit/utils/test_store.py
```diff
@@ -29,11 +31,11 @@ def push_metadata(self, title: str, stac_content: dict, stac_type: str):  # prag
     def retrieve_metadata(self, title: str, stac_type: str):  # pragma NO COVER
         raise NotImplementedError
 
-    def write_metadata_only(self, attributes: dict):  # pragma NO COVER
+    def update_v3_metadata(self, attributes: dict):  # pragma NO COVER
```
ohsqueezy, line 34: Isn't this function in the Metadata class, and not Store

ohsqueezy (commented):
The Store class previously wasn't meant to open arbitrary paths, so this would actually be a somewhat significant change (I can see how Store.mapper makes that a bit confusing - it probably should have been named Store._mapper to make it clear it wasn't meant to be used outside the class). I think there are a lot of pros for maintaining the previous behavior. For one, it makes it clear the store is a tool for accessing the data managed by this software, and not for general access. So it follows that testing becomes a lot simpler - we don't need to test what happens when a range of arbitrary inputs are opened, we just test what happens when the expected Zarr data is either detected or not detected.

One way to maintain the behavior would be to add functions to Store for specific files we need, like Store.zarr_json, Store.zmetadata, Store.zarray.

I'm also thinking there should be checks in the v3 functions for whether a dataset with v3 data was actually submitted, and vice versa for the v2 functions. The actual functions could go in the Store class, and I guess there is already Store.has_v2_metadata, but I don't see an equivalent for Store.has_v3_metadata.

PR #210, opened 2025-05-09
Sophisticated S3 Error handling + improved time span management

rbanick:
…e error type

### DIFF
--- a/examples/managers/chirps.py
+++ b/examples/managers/chirps.py
@@ -47,7 +47,7 @@ class CHIRPS(DatasetManager):
         static_metadata = {
             "coordinate reference system": "EPSG:4326",
             "update cadence": self.update_cadence,
-            "temporal resolution": self.time_resolution.__str__(),
+            "temporal resolution": str(self.time_resolution),
             "spatial resolution": self.spatial_resolution,
             "spatial precision": 0.00001,
             "provider url": "http://chg.geog.ucsb.edu/",
--- a/examples/managers/chirps.py
+++ b/examples/managers/chirps.py
@@ -10,6 +10,7 @@ from abc import ABC
 
 from gridded_etl_tools.dataset_manager import DatasetManager
 from gridded_etl_tools.utils import extractor
+from gridded_etl_tools.utils.time import TimeSpan
 
 
 class CHIRPS(DatasetManager):
@@ -46,7 +47,7 @@ class CHIRPS(DatasetManager):
         static_metadata = {
             "coordinate reference system": "EPSG:4326",
             "update cadence": self.update_cadence,
-            "temporal resolution": self.time_resolution,
+            "temporal resolution": self.time_resolution.__str__(),
             "spatial resolution": self.spatial_resolution,
             "spatial precision": 0.00001,
             "provider url": "http://chg.geog.ucsb.edu/",
@@ -94,7 +95,7 @@ class CHIRPS(DatasetManager):
     collection_name = "CHIRPS"
     """Overall collection of data. Used for filling and referencing STAC Catalog."""
 
-    time_resolution = DatasetManager.SPAN_DAILY
+    time_resolution = TimeSpan.SPAN_DAILY
     """Increment size along the "time" coordinate axis"""
 
     data_var = "precip"
--- a/gridded_etl_tools/utils/attributes.py
+++ b/gridded_etl_tools/utils/attributes.py
@@ -325,14 +325,16 @@ class Attributes(ABC):
     minimum and maximum permissible values for common units
     """
 
-    ignorable_extraction_errors: tuple = ()
+    ignorable_extraction_errors: tuple[Exception] = ()
     """
-    A tuple of errors that are ignorable during S3 extraction.
+    A tuple of errors that are ignorable during S3 extraction and should trigger skipping the extract
+    entirely and moving to the next file.
     """
 
-    unsupported_extraction_errors: tuple = ()
+    unsupported_extraction_errors: tuple[Exception] = ()
     """
-    A tuple of errors that are unsupported during S3 extraction.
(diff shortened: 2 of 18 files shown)


gridded_etl_tools/utils/attributes.py
```diff
@@ -325,6 +325,16 @@ def store(self, new_store):
     minimum and maximum permissible values for common units
     """
 
+    ignorable_extraction_errors: tuple = ()
```
ohsqueezy, line 328: Minor thing but `()`  gets typed as tuple automatically, but actually if possible it would be good to declare the contained type like `tuple[Exception]`

gridded_etl_tools/utils/extractor.py
```diff
@@ -326,7 +326,7 @@ class S3Extractor(Extractor):
     be added to the given `DatasetManager`'s list of Zarr JSONs at `DatasetManager.zarr_jsons`.
     """
 
-    def __init__(self, dm: dataset_manager.DatasetManager):
+    def __init__(self, dm: dataset_manager.DatasetManager, **kwargs):
```
ohsqueezy, line 329: We should explicitly add `concurrency_limit` as an argument (with a default of 8) and leave out `**kwargs` (same as HTTPExtractor). I think I just missed including concurrency_limit initially.

gridded_etl_tools/utils/extractor.py
```diff
@@ -335,8 +335,12 @@ def __init__(self, dm: dataset_manager.DatasetManager):
         dm
             Source data for this dataset manager will be extracted
         """
-        super().__init__(dm)
+        super().__init__(dm, **kwargs)
         self.dm.zarr_jsons = []
+        if not isinstance(self.dm.ignorable_extraction_errors, tuple):
```
ohsqueezy, line 340: I have a few thoughts on this structurally.

Most importantly, I don't think this is the place to check whether these tuples are correct. If DM is created with a bad value for extraction_errors, it should get caught at creation.

However, I also think the errors are better implemented as arguments to the S3Extractor constructor and therefore as attributes of S3Extractor and not as attributes of DM. Basically, Extractor needs to know about them, and DM doesn't. They could even simply be arguments to S3Extractor.request.

Lastly, the "Pythonic" way to handle checking the type would be to allow lists and automatically convert a list into a tuple. It's also common to allow a single argument, which would be converted into a tuple.

gridded_etl_tools/utils/time.py
```diff
@@ -0,0 +1,205 @@
...
+            If the value is not positive
+        ValueError
+            If the unit is not one of the valid time units
+        """
+        if self.value <= 0:
+            raise ValueError(f"Time unit value must be positive, got {self.value}")
+        if self.unit not in self._CONVERSION_FACTORS:
+            raise ValueError(f"Invalid time unit: {self.unit}. Must be one of {list(self._CONVERSION_FACTORS.keys())}")
+
+    def to_minutes(self) -> int:
```
ohsqueezy, line 56: I'm concerned about the approximations. Would we ever use them, and should we be using them if we do?

gridded_etl_tools/utils/time.py
```diff
@@ -0,0 +1,205 @@
...
+        str
+            A string in the format "{value} {unit}"
+        """
+        return f"{self.value} {self.unit}"
+
+
+T = TypeVar("T", bound="TimeSpan")
+
+
+class TimeSpan(Enum):
```
ohsqueezy, line 87: I'm not sure what the scope is for these utility classes but I will say a nice thing for a TimeSpan would be if it could be converted into a datetime.timedelta. That way you can add/subtract a TimeSpan from a datetime.

rbanick: I like it

ohsqueezy (commented):
For the exception checking, my thoughts are that in general, exceptions aren't really passed around as arguments, outside of meta-programming at least. What are the exceptions we're planning to add to the lists? If it's FileNotFoundError and ValueError, like in the tests, we could just add an argument like skip_missing_files to S3Extractor.request.

General feedback about the time classes: I think better time and date handling is very welcome, but my concern is that it's on the one hand complex for us to test and maintain, and on the other hand not sophisticated enough to be dropped in anywhere and interact nicely with other time libraries (time, datetime, pandas, etc). Ideally, there would be a good Python module, but I admit I don't know of any on PyPI or anything. So I'll just leave that at general thoughts, and maybe we can just roll with it.

examples/managers/chirps.py
```diff
@@ -46,7 +47,7 @@ def static_metadata(self):
         static_metadata = {
             "coordinate reference system": "EPSG:4326",
             "update cadence": self.update_cadence,
-            "temporal resolution": self.time_resolution,
+            "temporal resolution": self.time_resolution.__str__(),
```
eschechter, line 50: you can call `str(self.time_resolution)` instead

eschechter (approved):

gridded_etl_tools/utils/extractor.py
```diff
@@ -326,18 +326,37 @@ class S3Extractor(Extractor):
...
     """
 
-    def __init__(self, dm: dataset_manager.DatasetManager):
+    def __init__(
+        self,
+        dm: dataset_manager.DatasetManager,
+        concurrency_limit: int = 8,
+        ignorable_extraction_errors: list[type[Exception]] | tuple[type[Exception], ...] = (),
+        unsupported_extraction_errors: list[type[Exception]] | tuple[type[Exception], ...] = (),
+        **kwargs,
```
ohsqueezy, line 335: Changes look good but I don't think **kwargs is needed (Extractor only accepts dm and concurrency_limit)

ohsqueezy (approved):

PR #213, opened 2025-05-15
allow dict and None attrs to be written

eschechter:


rbanick (approved):
Can you create a "regression" test proving that JSON insertion into a Zarr's attributes is fine? Perhaps unnecessary, just my paranoia going off

rbanick (approved):

PR #226, opened 2025-07-07
Flexible TimeSpan creation

rbanick:


gridded_etl_tools/utils/time.py
```diff
@@ -91,157 +91,135 @@ def __str__(self) -> str:
...
+            "half_hourly": cls.create("minutes", 30),
+            "hourly": cls.create("hours", 1),
+            "3hourly": cls.create("hours", 3),
+            "6hourly": cls.create("hours", 6),
+            "daily": cls.create("days", 1),
+            "weekly": cls.create("weeks", 1),
+            "monthly": cls.create("months", 1),
+            "yearly": cls.create("years", 1),
+            "seasonal": cls.create("seasons", 1),
         }
```
eschechter, line 161: This could equivalently be
```python
predefined_map = {value: cls(key) for key, value in cls._PREDEFINED_STRINGS.items()}
```

eschechter (commented):
I also think we should avoid having `months`, `years`, and `seasons` in `TimeUnit._CONVERSION_FACTORS`, since they're unused and could lead a user to believe that we're treating them as fixed spans, which we're not. I think we could remove them from the dict and instead use
```python
    if self.unit not in get_args(TimeUnitType):
        raise ValueError(f"Invalid time unit: {self.unit}. Must be one of {get_args(TimeUnitType)")
```
for validation, where `get_args` is in the builtin `typing` module.

PR #265, opened 2026-04-22
fix: Publish update guard

rbanick:


gridded_etl_tools/utils/publish.py
```diff
@@ -350,6 +355,25 @@ def update_zarr(self, publish_dataset: xr.Dataset):
         else:
             self.info("No new records to append to original zarr")
 
+    def _check_for_incomplete_write(self, original_dataset: xr.Dataset | None):
+        """
+        Raise ConcurrentWriteError if the existing Zarr's update_in_progress flag is True, which indicates
```
ohsqueezy, line 360: As far as I know, this isn't the case anymore. We implemented the `finally` clause in to_zarr so the update_in_progress flag would be almost guaranteed to be reset even when an error happens. I think the exception would be a scenario where the finally clause also fails because `write_metadata_only` fails.

So I think the situation is more like what you were describing on Slack where update_in_progress is literally true.

gridded_etl_tools/utils/publish.py
```diff
@@ -327,6 +331,7 @@ def update_zarr(self, publish_dataset: xr.Dataset):
             A dataset containing all updated (insert) and new (append) records
         """
         original_dataset = self.store.dataset()
+        self._check_for_incomplete_write(original_dataset)
```
ohsqueezy: I think Claude is right that this would never run with no previous dataset, so we don't need to allow None as a parameter. Maybe the best solution is to just check if original_dataset is None before running this.

gridded_etl_tools/utils/publish.py
```diff
@@ -350,6 +355,25 @@ def update_zarr(self, publish_dataset: xr.Dataset):
         else:
             self.info("No new records to append to original zarr")
 
+    def _check_for_incomplete_write(self, original_dataset: xr.Dataset | None):
```
ohsqueezy: Maybe I'm wrong but I feel like conventionally "checking" implies the function will return a value, but a function that runs just to conditionally raise an exception is called "raise_if..."

gridded_etl_tools/utils/publish.py
rbanick: Yeah the comments / docstrings are wrong, I will fix them

gridded_etl_tools/utils/publish.py
rbanick: I think we can get ride of `None`. This is running in `update_zarr`, if we're somehow there without an original dataset we have bigger problems

gridded_etl_tools/utils/publish.py
```diff
@@ -350,6 +355,24 @@ def update_zarr(self, publish_dataset: xr.Dataset):
...
 
+    def _raise_if_concurrent_write(self, original_dataset: xr.Dataset):
+        """
+        Raise ConcurrentWriteError if the existing Zarr's update_in_progress flag is True, which indicates
+        an ongoing write from elsewhere. This prevents a double-append which would leave the on-disk
+        time coordinate and the in-memory view of it out of sync.
+
+        Parameters
+        ----------
+        original_dataset : xr.Dataset | None
```
ohsqueezy: Don't forget to remove None here too

gridded_etl_tools/utils/publish.py
```diff
@@ -350,6 +355,24 @@ def update_zarr(self, publish_dataset: xr.Dataset):
...
+            The existing dataset as opened from the store, or None if no Zarr exists yet
+        """
+        if original_dataset.attrs.get("update_in_progress") is True:
+            raise ConcurrentWriteError(
+                "The existing Zarr's 'update_in_progress' flag is True, indicating a prior write did not "
```
ohsqueezy: I think the message is a little misleading. Actually it's kind of dangerous because what if an update is running and you do either of the suggested fixes :) What's indicated is an update is in progress, not that an update didn't complete cleanly. A good follow up message though could be to make sure a previous update is actually running and didn't crash and break the flag.

gridded_etl_tools/utils/publish.py
```diff
@@ -350,6 +355,25 @@ def update_zarr(self, publish_dataset: xr.Dataset):
...
+        Parameters
+        ----------
+        original_dataset : xr.Dataset
+            The existing dataset as opened from the store
+        """
+        if original_dataset.attrs.get("update_in_progress") is True:
+            raise ConcurrentWriteError(
+                "The existing Zarr's 'update_in_progress' flag is True, indicating a concurrent update "
+                "is in progress. Check whether there is an ongoing, concurrent update to this dataset running "
+                "elsewhere. If not, the dataset is corrupted and you should consdier rolling back to a known-good "
```
ohsqueezy: Nice just check the minor spelling typo

gridded_etl_tools/utils/publish.py
```diff
@@ -27,6 +27,10 @@ class ZarrOutputError(Exception):
     """Raise when an exception occurs while the Zarr is being written"""
 
 
+class ConcurrentWriteError(Exception):
+    """Raise when a Zarr's update_in_progress flag indicates a prior write did not complete cleanly"""
```
ohsqueezy: Might want to update this docstring as well

ohsqueezy (approved):

PR #266, opened 2026-05-04
feat: Tweak version attrs

rbanick:


ohsqueezy (commented):
I think the best thing for this and the managers PR would be to use the [packaging](https://packaging.pypa.io/en/latest/version.html#packaging.version.Version) module.

Otherwise, there are a bunch of subtle annoyances. For example, the module can compare my_chirps.version to another_chirps.version, or to a literal like Version("2") or Version("3").

rbanick:
@ohsqueezy this is really meant to prepare metadata we can ingest into the registry and/or custom catalogs. So custom Packaging objects would complicate encoding versions into a JSON and anyways be overkill.

That said I'm open to more rigorous treatment that's JSON compatible. Thoughts?

ohsqueezy:
I think using plain `str` as the type and then doing the comparison with `packaging` would be good. The version strings have to be a certain format but they're pretty flexible. You can include arbitrary strings using the `+` character for example.

```
# Stored in JSON as "3.14+satellite"
> Version("3.14+satellite") == Version("3.14+drone")
False
> Version("3.14+satellite") < Version("3.15")
True
> Version("3.14+satellite") < Version("3.15+satellite")
True
```

ohsqueezy (approved):

PR #267, opened 2026-05-05
implement uv.lock and rewire CI/CD to use it

rbanick:


pyproject.toml
```diff
@@ -103,3 +102,6 @@ namespaces = false
 ignore = ["E203", "E701", "W503"]
 exclude = ["temp/*", "build/*", ".nox", ".venv"]
 max-line-length = 120
+
```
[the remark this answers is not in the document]
rbanick: Nope, this works fine with 0.11.x of uv

.github/workflows/tests.yml
```diff
@@ -0,0 +1,67 @@
...
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.TESTING_AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.TESTING_AWS_SECRET_ACCESS_KEY }}
+        run: |
+          uv run pytest \
+            --log-level=INFO \
+            --log-cli-level=INFO \
+            --cov=gridded_etl_tools \
+            --cov-append \
+            --cov-config=.coveragerc \
```
arbolabid, line 62: using .coveragerc is the default i believe

.github/workflows/tests.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Run system tests
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.TESTING_AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.TESTING_AWS_SECRET_ACCESS_KEY }}
+        run: |
+          uv run pytest \
+            --log-level=INFO \
+            --log-cli-level=INFO \
+            --cov=gridded_etl_tools \
+            --cov-append \
```
arbolabid, line 61: can probably be removed since this is a single fresh run in an action

pyproject.toml
```diff
@@ -64,42 +60,33 @@ dependencies = [
 [project.urls]
 Homepage = "https://arbol.io/"
 
-[project.optional-dependencies]
-testing = [
+[dependency-groups]
+dev = [
+    "black",
+    "flake8",
+    "flake8-pyproject",
```
arbolabid, line 67: In other PRs, this is capitalized to Flake8-pyproject vs lowercased here, should pick one style for consistency. The lowercase version is probably preferable

arbolabid (approved):
comments on other PRs also apply

.github/workflows/tests.yml
rbanick: This actually specifically addresses a bug w/ coverage. We intentionally run two separate pytest processes in sequence — unit tests write .coverage, then system tests need --cov-append to accumulate into it rather than overwrite it. Without this flag, the system test step would erase the unit coverage and the final coverage report would only see system test data, giving ~61% instead of 100%.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,69 @@
...
+        uses: awalsh128/cache-apt-pkgs-action@acb598e5ddbc6f68a970c5da0688d2f3a9f04d05  # v1.6.0
+        with:
+          packages: cdo netcdf-bin libeccodes0
+          version: 1.0
+      - name: Refresh dynamic linker cache
+        run: sudo ldconfig
+
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+        with:
+          fetch-depth: 0
```
arbolabid:
```suggestion
```

.github/workflows/tests.yml
arbolabid: ahhhh, ok yeah definitely keep

arbolabid (approved):
🚀

PR #272, opened 2026-06-02
feat: Support monthly datasets in calculate_update_time_ranges

rbanick:
Calendar units (months, years) have no fixed pd.Timedelta, so the prior gap detection silently returned wrong ranges. Branch the contiguity check on time unit: fixed-length units keep the Timedelta comparison, while calendar units measure gaps in whole calendar months via a year*12+month ordinal. Reject cftime-backed calendar datasets with a descriptive NotImplementedError; seasons remain intentionally unsupported.

Add monthly fixtures, a monthly manager, and tests for the monthly and cftime-rejection paths.


gridded_etl_tools/utils/publish.py
```diff
@@ -588,22 +588,53 @@ def calculate_update_time_ranges(
...
+                )
+            # Collapse each timestamp to an absolute month ordinal (year * 12 + month). Differences in this
+            # space are exact whole-month counts, sidestepping variable month/year lengths entirely.
+            spacing = complete_time_series.dt.year * 12 + complete_time_series.dt.month
+            step = {"months": 1, "years": 12}[time_unit.unit] * time_unit.value
+            # `inf` plays the same role for the integer ordinals that pd.Timedelta.max plays below: it
+            # forces the first/last elements (whose neighbor diff is NA) to register as range endpoints.
+            fill = float("inf")
+        else:
+            _PANDAS_TIMEDELTA_ALIAS = {"minutes": "min", "hours": "h", "days": "D", "weeks": "W"}
```
ohsqueezy: This would be a good property to add to TimeUnit. Then you could just call time_unit.pandas_alias.

gridded_etl_tools/utils/publish.py
```diff
@@ -588,22 +588,53 @@ def calculate_update_time_ranges(
...
+                raise NotImplementedError(
+                    f"calculate_update_time_ranges received non-datetime64 time values "
+                    f"(dtype '{complete_time_series.dtype}') for a '{time_unit.unit}' dataset. This most "
+                    "likely indicates a cftime calendar (e.g. 360_day, noleap), which is not yet supported "
+                    "for calendar-unit time ranges. Convert the time coordinate to datetime64 first."
+                )
+            # Collapse each timestamp to an absolute month ordinal (year * 12 + month). Differences in this
+            # space are exact whole-month counts, sidestepping variable month/year lengths entirely.
+            spacing = complete_time_series.dt.year * 12 + complete_time_series.dt.month
+            step = {"months": 1, "years": 12}[time_unit.unit] * time_unit.value
```
ohsqueezy: Not sure if Claude is trying to show off but I think most would agree a more readable alternative is better
```
if time_unit == "years":
    step = 12 * time_unit.value
else:
    step = time_unit.value
```
or
```
step = time_unit.value
if time_unit == "years":
    step *= 12
```
etc

ohsqueezy (approved):
The algorithm is too complex for me to follow line by line once the range starts getting built but it's easily tested and the test looks thorough so it's fine with me

